PR #3, opened 2017-01-11
Adds config params

### LINKED ISSUE #2: cannot provide options
there is no way to provide options in config and seems it doesn't do anything to my js code

colek42:
I had to add the .vscode/settings.json to the git ignore.  Mine were completely different.  This is also my first time working on an extension so things can probably be done better.

Fixes #2


esbenp:
This is awesome, thanks for contributing! I have two concerns: 

1. I do not think you can remove the `.vscode/settings.json`? What issue did you have with it?
2. Maybe it would be better if the configuration defaults were all undefined? So that we use `prettier`'s defaults. I.e. if `prettier` change default `printWidth` from `80` to `120` it would still be `80` for people using this extension. Would it not be better if we follow defaults set by `prettier`?

colek42:
1.  I think the settings .json is your specific workspace settings.
2.  I am fine either way, though when they are defined it gives context in the settings file.  What are the normal conventions?

image

esbenp:
Looking at other plugins they keep the `settings.json`
https://github.com/Microsoft/vscode-go/blob/master/.vscode/settings.json
https://github.com/felixfbecker/vscode-php-debug/blob/master/.vscode/settings.json
https://github.com/Dickurt/vscode-php-formatter/blob/master/.vscode/settings.json

And it is also part of the boilerplate for new libraries so I am pretty sure it is supposed to be there.

I think you are right about the default values. Let us keep them for now. I have been looking around and seems like no one uses `undefined` as a default value.

esbenp:
For the discussion about default parameters, the Atom plugin also just copies them over. https://github.com/jlongster/prettier-atom/pull/9/files 

So I think your solution is fine 👍  :-)

colek42:
Should be good to go.  Probably want to do a merge sqaush due to my abuse of commits here

esbenp:
Amazing stuff @colek42

colek42:
Thanks.  I'll probably hack away on this some more.  I'll add an issue and let you know I am working on it to prevent duplication of work.

PR #18, opened 2017-01-15
[Breaking Change] Use registerDocumentRangeFormattingEditProvider

CiGit:
This uses vscode's formatter API `languages.registerDocumentRangeFormattingEditProvider`.
It replaces vscode's JavaScript formatter with prettier.

➕  Less options/keybindings required.

➖  The original formatter is no more accessible if you still want it.

It will run on following [commands](https://code.visualstudio.com/docs/customization/keybindings#_rich-languages-editing):
- `editor.action.formatDocument` (default key: `Shift+Alt+F`)
- `editor.action.formatSelection` (default key: `Ctrl+K Ctrl+F`)
- if `editor.formatOnSave` is enabled

Added javascriptreact lang support

The original code was not removed, I think a discussion is clearly needed as it would be a **breaking change**.

So choices with this PR are:
 - extension command (`prettier.format`) (reject this PR)
 - formatter API (I update this PR: remove command)
 - Both, IMO a bad option... (_this_)

Note: Missing an updated readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,21 +31,10 @@
   ],
   "main": "./out/src/extension",
   "contributes": {
-    "commands": [
-      {
-        "command": "prettier.format",
-        "title": "Prettier: Format Javascript"
-      }
-    ],
     "configuration": {
         "type": "object",
         "title": "Prettier - JavaScript formatter configuration",
         "properties": {
-            "prettier.formatOnSave": {
-                "type": "boolean",
-                "default": false,
-                "description": "Format on save"
-            },
             "prettier.printWidth": {
                 "type": "integer",
                 "default": 80,
--- a/package.json
+++ b/package.json
@@ -26,8 +26,7 @@
   ],
   "activationEvents": [
     "onLanguage:javascript",
-    "onLanguage:javascriptreact",
-    "onCommand:prettier.format"
+    "onLanguage:javascriptreact"
   ],
   "main": "./out/src/extension",
   "contributes": {
--- a/package.json
+++ b/package.json
@@ -11,12 +11,12 @@
   },
   "homepage": "https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode",
   "repository": {
-      "type": "git",
-      "url": "https://github.com/esbenp/prettier-vscode.git"
+    "type": "git",
+    "url": "https://github.com/esbenp/prettier-vscode.git"
   },
   "license": "MIT",
   "bugs": {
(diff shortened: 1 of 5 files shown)


CiGit:
Here are some new changes.
* Removed `Prettier` action.

* deprecate `useFlowParser` for the more general `parser` setting
  > Introduce --parser/parser option and deprecate --flow-parser/useFlowParser (jlongster/prettier#342)

* Update Readme and Changelog
* Rebased on master

IMO this is how the extension should look like.
Feel free to ask for changes or tell me if you do not like it. I won't mind !

esbenp:
Hey. I am terribly sorry I have not gotten back to you on this! I will review this weekend, I promise.

package.json
```diff
@@ -26,56 +26,55 @@
   ],
   "activationEvents": [
     "onLanguage:javascript",
+    "onLanguage:javascriptreact",
     "onCommand:prettier.format"
```
esbenp: Should this one not be removed as well?

CiGit: Ow, I missed that one.

CiGit:
@esbenp I don't know what may be wrong. Are you running vscode 1.8.1 ?
I may also provide a [DocumentFormattingEditProvider](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-namedocumentformattingeditprovideraspan-classcodeitem-id612documentformattingeditproviderspan) to test your case.
In my memory, Document Formatting calls the range formater with full range (can't remember where I've seen this)

esbenp:
Yes I am running 1.8.1.
It does run some sort of formatting which I am guessing is the built in formatting? However it is not running Prettier.

CiGit:
Correction done. Also removed `useFlowParser` from the underlying lib

There is a builtin formater provided by typescript server. `registerDocumentRangeFormattingEditProvider` should replace this default formater.
I've tried on Linux and Windows 10 machines. Working in both cases ...

No other extension providing a JS formater installed ?

I'll try to also `registerDocumentFormattingEditProvider` a little later.

CiGit:
Hopefully this change will fix the troubles you have (others maybe too). I've not found my previous assertion (was it a dream: always calls range formater  instead of document if the first one is provided) so I think this should be good.

Separating both formatting options may help correct the output in the range case. Prettier is meant for a whole doc. But that's an other future issue. I mostly think about the `EOL` at the end

esbenp:
Looks very good :-) Thank you so much for this contribution!

Issue #11, opened 2017-01-12
Add key binding to prettify entire document


colek42:
ref: https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributeskeybindings

I should be able to get to this in the next couple days.  I'll post here when I start in case you beat me to it.

This should also be a config param.

lrowe:
I think #18 fixes this since Format Document has a default key binding (shift-option-F on Mac.)

PR #27, opened 2017-01-23
Add key binding rule to prettify focused text

ConAntonakos:
This contributes to #11 partially by defining a key binding for prettifying depending on the editor's focused text.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,12 @@
         "title": "Prettier: Format Javascript"
       }
     ],
+    "keybindings": [{
+      "command": "prettier.format",
+      "key": "ctrl+p",
+      "mac": "cmd+p",
+      "when": "editorTextFocus"
+    }], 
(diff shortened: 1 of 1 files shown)


esbenp:
Thanks for your contribution(s)! It might be at bit much to bind `CMD+p` to prettify, wont it? Since `CMD+p` opens Quick Open - at least on Mac?

ConAntonakos:
Thanks for the response, @esbenp. That's a good point. I started migrating over to VS Code from Sublime not too long ago, so still in the process of learning all the shortcuts and tricks. ;)

Do you have any suggestions or preferences?

CiGit:
Due to implementation change in latest release, there is a key binding by default [depending on the system ](https://code.visualstudio.com/docs/customization/keybindings#_keyboard-shortcuts-reference) for `Format Document`

PR #31, opened 2017-01-25
Make formatOnSave work only in javascript files

vasa-chi:
Currently, if `formatOnSave` is `true`, this plugin tries to format every file (.travis.yml, for instance).
This solves this problem.

Although #18 looks better for this sort of plugin, IMHO...

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,12 +20,16 @@ interface Rangeable {
 }
 
 export function activate(context: ExtensionContext) {
-    const eventDisposable = (workspace as any).onWillSaveTextDocument(e => {
+    const eventDisposable = workspace.onWillSaveTextDocument(e => {
         const document = e.document;
 
         if (!document.isDirty) {
             return;
         }
+
+        if (document.languageId !== 'javascript') {
+            return;
+        };
 
         const config: PrettierConfig = workspace.getConfiguration('prettier') as any;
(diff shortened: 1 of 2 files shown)


jaydp17:
Yup happened to me too

src/extension.ts
```diff
@@ -20,13 +20,17 @@ interface Rangeable {
 }
 
 export function activate(context: ExtensionContext) {
-    const eventDisposable = (workspace as any).onWillSaveTextDocument(e => {
+    const eventDisposable = workspace.onWillSaveTextDocument(e => {
```
esbenp: This will generate compilation error

```
src/extension.ts(23,39): error TS2339: Property 'onWillSaveTextDocument' does not exist on type 'typeof workspace'.
```

vasa-chi: I've updated the "engines": "vscode" to 1.8.0, where onWillSaveTextDocument is defined on workspace. You probably need to run `npm install` before compilation.

vasa-chi:
```
~/workspace/prettier-vscode(master*) » npm run vscode:prepublish 

> prettier-vscode@0.6.0 vscode:prepublish /Users/vasa/workspace/prettier-vscode
> tsc -p ./

~/workspace/prettier-vscode(master*) »
```

esbenp:
Close in favor of #18?

PR #36, opened 2017-02-07
Show error message to user

CiGit:
Instead of silently printing error in a console a user certainly won't
see, show an error message with an option to show it.
Scroll to and move cursor to it.

Handles multiple error in case a user saves all files.

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -1,19 +1,22 @@
 import {
     workspace,
+    window,
     DocumentRangeFormattingEditProvider,
     DocumentFormattingEditProvider,
     Range,
     TextDocument,
     FormattingOptions,
     CancellationToken,
-    TextEdit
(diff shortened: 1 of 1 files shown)


src/PrettierEditProvider.ts
```diff
@@ -1,26 +1,33 @@
...
     CancellationToken,
-    TextEdit
+    TextEdit,
+    Selection,
+    Position
 } from 'vscode';
 
 const prettier = require('prettier');
 
+type ShowAction = "Show";
```
CiGit: Was wondering if _Show_ was a good text for the action. I also had _Go to_ in mind.
Any preference ?

CiGit: Was wondering if _Show_ was a good text for the action. I also had _Go to_ in mind.
Any preference ?

CiGit:
rebased

PR #40, opened 2017-02-21
Allow disabling formatting when prettier is not installed in the proj…

jacobp100:
Allow disabling formatting when prettier is not installed in the project to assist with editor.saveOnFormat

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ OR
 
 #### Format On Save
 
-Respects `editor.formatOnSave` setting.
+Respects `editor.formatOnSave` setting. If you want to turn this on, but not to run in projects that do not have prettier installed, use `prettier.onlyFormatWhenInstalled: true`.
 
 ### Settings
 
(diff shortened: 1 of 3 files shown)


jacobp100:
Or is the way to go about this to use workspace settings? Not sure!

CiGit:
Without testing this I think this is also disabling the default formatter but I may be wrong.
Wouldn't it be easier to disable the extension for the current workspace?

jacobp100:
Yea, tbh I only realised you could do per-project settings after I coded this and was testing 😂 Fool free to close if you don't think it would be useful.

frederickfogerty:
I think this is still useful, as I might want to use another formatter, and just disable prettier. Using `editor.formatOnSave` disables all formatters.

CiGit:
I did not mean to disable `formatOnSave` but disable the extension for the workspace

frederickfogerty:
Oh, @CiGit how do you do that?

CiGit:
There is a cogwheel for each extension in the extensions pane.

Issue #29, opened 2017-01-24
Use prettier from current workspace


vasa-chi:
Currently plugin uses internal dependency on prettier. I think it would be better to use prettier from current projects node_modules, if it exists.

CiGit:
ref #33

esbenp:
Makes sense I guess. As @CiGit laid out in #33 there could be a prioritization order a long the lines of.

1. Use local version
2. Use global version
3. Use package version

As far as I can tell this is what is going on here https://github.com/Microsoft/vscode-eslint/blob/master/eslint-server/src/server.ts#L257

dburles:
Hey any update on this one @esbenp I was just bitten by this after updating to 0.18.0.

esbenp:
Honestly, February is pretty booked with work, so wont have time until at the end. Any PRs are welcome though..

CiGit:
I'll certainly try to work on this in these upcoming days.
@esbenp Is it okay for you if there is a new configuration option to define the global installation folder(/usr/lib/node_modules/prettier for example) or would you prefer relying on PATH to find the global executable cli and then mess around to find the lib.
I see the later more difficult to implement correctly. It depends on OS/package manager (npm, yarn, ...) combination

esbenp:
Looking at https://github.com/Microsoft/vscode-eslint/blob/master/eslint-server/src/server.ts#L349 it seems they way they do it is to use `Files` of `vscode-languageserver-node`. You can see the method here https://github.com/Microsoft/vscode-languageserver-node/blob/8291f55041ea023c4acefa73d8f25f5384aa6426/server/src/files.ts#L178 
Maybe just copy-paste `resolveGlobalNodePath` into our repo and use it to resolve the global node path in case no local is found? What do you think?

CiGit:
`resolveGlobalNodePath`, like `requireg`, are based on `npm config get prefix` which is only relevant to npm.
In prettier's readme, installation is **first** shown with `yarn global add prettier`.
But I get it, you prefer no additional config 😉 and I agree with that. I will just need to investigate some more.

esbenp:
Ah yeah I get what you mean now. Well, maybe there is nothing wrong with adding a path config and set it to `null` as default. If `null`, run the 1. local -> 2. npm global -> plugin. If not null, use that if executable is found or fall back to previous priority

CiGit:
Sorry if I was unclear.
If I get your idea correctly:
1. try local
[...]

PR #38, opened 2017-02-15
Update to prettier 0.21.0

CiGit:
* Handle `jsxBracketSameLine` introduced in prettier [0.17.0 ](https://github.com/jlongster/prettier/blob/master/CHANGELOG.md#0171)
  > Add option for putting > on the last line in jsx

* Update prettier to 0.21.0

* Change `trailingComma` from boolean value to string in `['none', 'es5', 'all']`

* Correct bracket spacing description in configuration (My bad, I forget this file in #37)

collides with #45

PR #45, opened 2017-02-26
update prettier to 0.19.0

dburles:
until #29 is answered the best solution is to keep the dependency up to date

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,6 @@
     "@types/mocha": "^2.2.32"
   },
   "dependencies": {
-    "prettier": "0.17.0"
+    "prettier": "0.19.0"
   }
 }


CiGit:
Thanks, could you also update `trailingComma` ? Option changed in 0.19.0 from a boolean value to string value in `['none', 'es5', 'all']` https://github.com/prettier/prettier/blob/master/CHANGELOG.md#0190

dburles:
Sure thing

CiGit:
Forgot to point you relevant parts.
Source files which change can be found as an example in #38.

dburles:
Thanks, would you mind pointing me in the direction of how to test the extension locally? first time working on a vscode extension.

CiGit:
Sure.
We have no automated tests. Maybe one day...
You can run the extension `F5` and then in the new window open a JS project to play with. You can also put breakpoints to have a *chrome like* debugger.

CiGit:
@dburles I've just done this, hope you won't mind if you were doing it #38

dburles:
I haven't had a chance to work on it yet so that's great!

Issue #72, opened 2017-04-13
add `semi` option


yukulele:
prettier recently add option to remove `semi` option
https://github.com/prettier/prettier#api

michrome:
+1, please support `semi`.

esbenp:
I just opened #73 
Can you gents let me know when prettier tags a release with the `semi` option?

michrome:
Will do! Thanks.

yukulele:
https://github.com/prettier/prettier/releases/tag/1.0.0

PR #78, opened 2017-04-14
Add semi and useTabs options, #72

julienp:


julienp:
Should have checked open request before :) I'll close this in favour of the already existing pull requests!

Issue #56, opened 2017-03-29
JSX files don't have the Format Document option


jshamley:
Thanks for an awesome plugin. This is working great for .js, .html and .json files but I do not see the option to format a .jsx document. I installed the extension through the marketplace and it is currently at version 0.11.0

I have this JSX extension installed with version 0.0.1 https://marketplace.visualstudio.com/items?itemName=TwentyChung.jsx

![screen shot 2017-03-29 at 9 11 00 am](https://cloud.githubusercontent.com/assets/1050828/24462176/92e554c6-1460-11e7-92ae-82b6b62df8aa.png)

esbenp:
What does it say in your status bar when you have a JSX file open? In the bottom of the window

<img width="415" alt="screen shot 2017-03-29 at 22 01 51" src="https://cloud.githubusercontent.com/assets/934819/24473968/60070930-14cb-11e7-84b6-c6e91aee57c3.png">

jshamley:
it says JSX instead of JavaScript React. 

![screen shot 2017-03-29 at 2 05 47 pm](https://cloud.githubusercontent.com/assets/1050828/24474142/e48d8794-1488-11e7-9f21-fdfa6c371c46.png)

esbenp:
Okay, so I added an activation event now that activates the plugin when the language is jsx. That should work. Give it a go and let me know :-)

jshamley:
Hmmm, still nothing. I updated and even restarted my machine to see if it would work. 

I notice that it shows the Format Document option in HTML, even though you don't have it listed in your `activationEvents` array.

![screen shot 2017-03-29 at 2 26 18 pm](https://cloud.githubusercontent.com/assets/1050828/24475004/cd6e73f4-148b-11e7-829f-7bfd53d6a98d.png)

CiGit:
@esbenp only the activation event is not enough. You need to also register the formatter [here](https://github.com/esbenp/prettier-vscode/blob/418ae37a055c815b8fb60148f9f6614d2e85df28/src/extension.ts#L10)

I've no time to take a deep look at the linked extension, but just with the readme, i don't see it's usefulness anymore. JSX is well supported out of the box by code, so I don't see the point in adding a new *language*.

CiGit:
I don't blame the extension. But it seems to be an old one developed during code's first days.
It has a lot of downloads tho

CiGit:
I forgot to respond to the initial post. 
[...]

PR #82, opened 2017-04-17
Register JSX files with the formatter

esbenp:
Seee #56


CiGit:
Is seems right.
But are we sure `jsx` is the language name?

I'm not convinced the [extension](https://marketplace.visualstudio.com/items?itemName=TwentyChung.jsx) is still useful. (reading comments and last update time 😕 )

I'll try to take a moment to dig into jsx extension's source (which I haven't found yet)

CiGit:
image

`jsx` is correct

PR #92, opened 2017-04-23
Eslint integration

### LINKED ISSUE #70: Add prettier-eslint integration
[The atom plugin](https://github.com/prettier/prettier-atom) has [added this](https://github.com/prettier/prettier-atom/blob/57534b166b0896a259df5324632c3be91bc4c20f/src/executePrettier.js#L30-L44) (maybe @robwise can give some tips). This way you and @RobinMalfait (who created [prettier-eslint-vscode](https://github.com/RobinMalfait/prettier-eslint-code)) could join forces to make this package even better!

CiGit:
Use prettier-eslint
new option: prettier.eslintIntegration (boolean)

## code refactor
making a type definition file.
requirePrettier is now requirePackage(name)

fixes #70


src/PrettierEditProvider.ts
```diff
@@ -65,7 +56,18 @@ function format(text: string, path: string): string {
         parser: parser,
         semi: config.semi,
         useTabs: config.useTabs,
-    });
+    };
+    if (config.eslintIntegration) {
+        const prettierEslint = requireLocalPkg(path, 'prettier-eslint') as PrettierEslintFormat;
```
CiGit: Is this really useful ?
prettier-eslint already loads what's in the workspace. And thus we could avoid some troubles with people having an older version (< 6.1.0) without `fallbackPrettierOptions`

CiGit:
Removed prettier-eslint project resolution.
As discussed with @kentcdodds in #70 it seems pointless and should avoid us some issues (doesn't work as stated because fallbackPrettierOptions is missing in local version)

ls-guillaume-lambert:
Do you need a review for this or it's ready to be merged? If yes, is there any guidelines I could follow to review? First time I'm looking at vscode/Typescript code 😉

ls-guillaume-lambert:
Really starting to like VSCode, looking forward for this to be merged to have basically the same setup I have on Atom 😄  Thanks for all your work on this too @kentcdodds !

CiGit:
Thanks for your interests ;-)
To be honest, I'm quite confident with the functionality. Where I'm a lot less is with the documentation:
- Should I add something more to explain what it does? Readme etc.
- My english? (I'm not a native english and maybe wrote a nonsense :smile: )

@ls-guillaume-lambert You can install that version locally and play with it, to increase *confidence*. We lack test ... :fire: (nicely said, we have none)
I'll give you a better guideline in about 30 minutes.

CiGit:
Seems you don't know vscode extensions / typescript.
### Typescript
You may start by some [typescript](https://www.typescriptlang.org/docs/tutorial.html):
If you know flow, it isn't really different at least for what we use in this extension. `something: sometype`
Except casting: `someting as sometype` and `Promise` is `Thenable`
Typescript gives us nice autocomplete/ typechecking

### VsCode
Documentation for extensions starts [here](https://code.visualstudio.com/docs/extensionAPI/overview). There is no point in reading everything. Typescript inside vscode will give you doc on various api.

### Develop / debug
1. Open this extension in vscode
1. `npm install`
1. Launch extension 
image
1. You can then set breakpoints and more (chromium debugger)

### Extension
I'll suppose you are on this branch. And hope we have enough comments in the code to be clear :-)
#### extension.ts
Entry point. Registers out formatter for different language id (javascript, javascriptreact) and checks for deprecated options.
#### types.d.ts
Defining typescript definitions for what we use (prettier and prettier-eslint mostly)
#### PrettierEditProvider.ts
The formatter. Most logic is here.
- Reading config
- Delegation to prettier or prettier-eslint
- Handling errors
#### requirePkg.ts
Function to load a *package* (only used for prettier) being in a package.json along formatted file's path.
Fallback to bundled *package* if none found.
prettier-eslint is always the bundled one.

CiGit:
@kentcdodds Thanks for the review.
Not knowing TS or vscode's specific thing is not such a problem. It loads 😆 . The rest is javascript logic with some type annotations.
I find it always preferable to have someone review a code change.

README.md
```diff
@@ -30,6 +30,9 @@ OR
 Respects `editor.formatOnSave` setting.
 
 ### Settings
+#### prettier.eslintIntegration (default: true)
```
ls-guillaume-lambert: Isn't the default `false` here? https://github.com/esbenp/prettier-vscode/pull/92/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40

CiGit: Right ;-)

ls-guillaume-lambert:
I'm getting some `.flowconfig` error when opening a project and running a file through `Format Document` with this PR's branch in VSCode:

```
Could not find a .flowconfig in . or any of its parent directories.
See "flow in... (length: 106)
```

And in the Debug Console:

```
rootPath null
```

Am I debugging this the good way?

1. Clone repo
2. Checkout this PR's branch
3. Open in VSCode
4. Go to `Debug` and hit the play icon beside `Launch Extension`
5. In the newly opened window, I add a folder project that has ESLint setup on it
6. Run the `Format Document` command (shouldn't it be `Prettier: format` instead?)
7. It works, but when I save I get the `.flowconfig` error (Is the `save` part why I'm getting errors, and I shouldn't save in the first place? 😉 )

Thanks for your time 😄

CiGit:
I highly doubt the error you have is from this extension. Seems to be a flow error.
Is there a `// @flow` at the top of your file ? 
Your setup seems right to me.
6. We replace the default formatter. This has some advantages (and some disadvantages). That's why it is bound to Format Document.

ls-guillaume-lambert:
Oh I see for the `Format Document`, that's OK. No I don't have `// @flow` anywhere in my project.

CiGit:
I can't help you much on that.
You have a debugger by the way ;-) (break on error)
I really think it's a flow extension. I don't see any other extension searching for a `.flowconfig`

CiGit:
I've found 
```
[2017-04-28T08:50:18.286Z] [INFO] nuclide - Path to Flow: flow
rootPath null
```
It's the debug output from *Flow Language Support* extension in my case. Disabling it removed the log.

ls-guillaume-lambert:
In any case I tried the `prettier.eslintIntegration` option and it worked perfectly :+1:

CiGit:
Thanks for looking at it.
Will release this soon.

ls-guillaume-lambert:
What's the next step until its released? Curious 😉

CiGit:
https://github.com/esbenp/prettier-vscode/releases/tag/v0.16.0

use `vsce` to build the package. (vsix file) and publish it.
In fact `vsce publish` does that in one user step.
Then wait some minutes till it shows up on the marketplace.

ls-guillaume-lambert:
Merci beaucoup! 👍

Issue #87, opened 2017-04-20
Vue support ?


jbrodriguez:
What I gather from this issue https://github.com/prettier/prettier/issues/651, is that vue support should be implemented in something like the editor.

prettier-atom had this PR https://github.com/prettier/prettier-atom/pull/44

Wondering if something similar can be done for vscode ?

CiGit:
It should be possible to do something. (maybe ugly, It seems vscode doesn't have embedded language support)
I don't use vue myself. Has vscode a language associated with it?

PR welcome ! :fire:

jbrodriguez:
> I don't use vue myself. Has vscode a language associated with it?

It has something called single file component (SFC). In a single file, you have html, script and styles sections.

So the idea is to single out the script (javascript/typescript) section and send it to prettier for formatting.

I'll try to take a look at the code and see how this can be done.

wuhaosky:
@jbrodriguez  I'll be waiting for your good news.

jsappme:
+1

muhajirframe:
+1

colwilson:
+1

CiGit:
Will support it once it lands in prettier. And it seems it will.

denisw:
Prettier has support for Vue components now (for the `<script>` and `<style>` parts): https://github.com/prettier/prettier/issues/2097

CiGit:
So does this extension. It's just disabled by default. `disabledLanguages`

PR #99, opened 2017-05-05
Implement vue support

jbrodriguez:
@CiGit,

This is mostly a proof-of-concept (as I don't have the extension dev env in my workstation), so it's also untested.

I didn't a find a way to avoid a direct check for a vue extension.

So this gets the document text, looks for the offsets of script tags, then converts them to a Range that will be processed by prettier.

Let me know what you think.

Related to #87

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -107,9 +107,28 @@ class PrettierEditProvider implements
         token: CancellationToken
     ): TextEdit[] {
         try {
+            const text = document.getText();
+            
+            if (document.fileName.slice(0, -4) === '.vue') {
+                const start = text.indexOf('<script>');
+                if (start !== -1) {
+                    const end = text.indexOf('</script>');
+                    if (end !== -1) {
+                        const startPos = document.positionAt(start + 8); // 8 = length of '<script>'
+                        const endPos = document.positionAt(end - 1); // -1 = character before '</script>'
+                        const range = new Range(startPos, endPos);
+
+                        return [TextEdit.replace(
+                            range,
+                            format(document.getText(range), document.fileName)
+                        )];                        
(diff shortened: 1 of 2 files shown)


CiGit:
I was thinking about it. What I would like when I format a vue file or anything with mixed  content (html, ...) I'd like to format the whole document not only a part of it.

In this PR registering a formater for vue means you can't register another one for this language, only one of them will run. (This is how it work in vscode)

Therefor I think it would be better to create a new extension for it (with support for every part) than doing it that way.

Another solution may be to implement that in [vetur](https://marketplace.visualstudio.com/items?itemName=octref.vetur) it already has options to format. And it formats the entire file. (It seems to be the must have extension for vue)
It may be worth to explore cross-extension calls (I know nothing about that). (Calling our `format` function from a complete extension like vetur)

src/PrettierEditProvider.ts
```diff
@@ -107,9 +107,28 @@ class PrettierEditProvider implements
         token: CancellationToken
     ): TextEdit[] {
         try {
+            const text = document.getText();
+            
+            if (document.fileName.slice(0, -4) === '.vue') {
```
CiGit: The file extension doesn't work when file is unsaved.
Should be `document.languageId === 'vue' `

I would also extract this next part into a function.

jbrodriguez:
You're right. I saw vetur [has plans to format using prettier](https://github.com/octref/vetur/issues/176#issuecomment-299499100).

I've been busy this week, will look into the changes you requested next week.

alexsasharegan:
+1 for working with Vetur. It's definitely the go-to vscode extension.

CiGit:
I've seen vue support in prettier. Not yet released. We will wait for that.

PR #100, opened 2017-05-07
Add TypeScript and CSS support and run code through prettier!

azz:
So as you may be aware, we're working pretty hard on on adding TypeScript support to prettier (prettier/prettier#13 and prettier/prettier#1480).

I thought now that we're parsing _almost_ everything, that it would be a good time to add some experimental support to this extension.

I've adding a new config option, `"prettier.runOnTypeScript"` that is not documented in `package.json`, so no-one will find it by accident. 

When the setting is enabled, the extension will also provide edits to TypeScript and TypeScript React files.

I then used this extension to run prettier on this extension's source code yodawg, with default settings plus `"prettier.tabWidth": 4`. I committed most of the changes.

Note that for this to work you need to install prettier@master from github, and install the correct `typescript-eslint-parser` commit from prettier's package.json.

When full TypeScript support is released in prettier, the only changed required are:

- [x] Remove the `"prettier.runOnTypeScript"` setting
- [x] Turn on TypeScript support if the installed version of prettier supports TypeScript (by version number).

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,12 +89,10 @@
           "type": "string",
           "enum": [
             "babylon",
-            "flow",
-            "typescript",
-            "postcss"
+            "flow"
           ],
           "default": "babylon",
-          "description": "Which parser to use. \"typescript\" and \"postcss\" are supported in prettier 1.4.0"
+          "description": "Override the parser. You shouldn't have to change this setting."
         },
         "prettier.semi": {
           "type": "boolean",
@@ -124,7 +122,7 @@
     "vscode": "^1.0.0"
   },
   "dependencies": {
-    "prettier": "1.4.0",
+    "prettier": "1.4.1",
     "prettier-eslint": "6.2.3",
     "read-pkg-up": "2.0.0",
     "semver": "^5.3.0"
--- a/package.json
+++ b/package.json
@@ -27,6 +27,8 @@
   "activationEvents": [
     "onLanguage:javascript",
     "onLanguage:javascriptreact",
+    "onLanguage:typescript",
+    "onLanguage:typescriptreact",
     "onLanguage:jsx"
   ],
   "main": "./out/src/extension",
@@ -109,7 +111,7 @@
     "test": "node ./node_modules/vscode/bin/test"
   },
   "devDependencies": {
-    "typescript": "^2.0.3",
+    "typescript": "~2.3.2",
     "vscode": "^1.0.0",
     "mocha": "^2.3.3",
     "@types/node": "^6.0.40",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,10 @@
     "onLanguage:javascriptreact",
     "onLanguage:typescript",
     "onLanguage:typescriptreact",
-    "onLanguage:jsx"
+    "onLanguage:jsx",
+    "onLanguage:css",
+    "onLanguage:less",
+    "onLanguage:scss"
   ],
   "main": "./out/src/extension",
   "contributes": {
@@ -86,10 +89,12 @@
           "type": "string",
           "enum": [
             "babylon",
-            "flow"
+            "flow",
+            "typescript",
+            "postcss"
           ],
           "default": "babylon",
-          "description": "Which parser to use. Valid options are 'flow' and 'babylon'"
+          "description": "Which parser to use. \"typescript\" and \"postcss\" are supported in prettier 1.4.0"
         },
         "prettier.semi": {
           "type": "boolean",
@@ -111,15 +116,17 @@
     "test": "node ./node_modules/vscode/bin/test"
   },
   "devDependencies": {
-    "typescript": "~2.3.2",
-    "vscode": "^1.0.0",
-    "mocha": "^2.3.3",
+    "@types/mocha": "^2.2.32",
     "@types/node": "^6.0.40",
-    "@types/mocha": "^2.2.32"
+    "@types/semver": "^5.3.31",
+    "mocha": "^2.3.3",
+    "typescript": "~2.3.2",
+    "vscode": "^1.0.0"
   },
   "dependencies": {
     "prettier": "1.4.0",
     "prettier-eslint": "6.2.3",
-    "read-pkg-up": "2.0.0"
+    "read-pkg-up": "2.0.0",
+    "semver": "^5.3.0"
   }
 }
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -9,20 +9,19 @@ import {
     CancellationToken,
     TextEdit,
     Selection,
-    Position
-} from "vscode";
+    Position,
+} from 'vscode';
 
-import { requireLocalPkg } from "./requirePkg";
+import { requireLocalPkg } from './requirePkg';
 
 import {
     PrettierVSCodeConfig,
     PrettierConfig,
     Prettier,
-    PrettierEslintFormat
-} from "./types.d";
-type ShowAction = "Show";
+    PrettierEslintFormat,
+} from './types.d';
 
-import { extname } from "path";
+type ShowAction = 'Show';
 
 /**
  * Format the given text with user's configuration.
@@ -30,7 +29,7 @@ import { extname } from "path";
  * @param path formatting file's path
  * @returns {string} formatted text
  */
-function format(text: string, path: string, customOptions: object): string {
+function format(text: string, { fileName, languageId}: TextDocument, customOptions: object): string {
     const config: PrettierVSCodeConfig = workspace.getConfiguration('prettier') as any;
     /*
     handle deprecated parser option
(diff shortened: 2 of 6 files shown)


src/PrettierEditProvider.ts
```diff
@@ -10,41 +10,49 @@ import {
...
 /**
  * Format the given text with user's configuration.
  * @param text Text to format
  * @param path formatting file's path
- * @returns {string} formatted text 
+ * @returns {string} formatted text
  */
 function format(text: string, path: string): string {
-    const config: PrettierVSCodeConfig = workspace.getConfiguration('prettier') as any;
+    const config: PrettierVSCodeConfig = workspace.getConfiguration(
```
azz: prettier/prettier#1547

CiGit:
Great!
I'm not sure in which case we need prettier@master ? To format this extension?

>Turn on TypeScript support if the installed version of prettier supports TypeScript (by version number).

This part seems to be complicated to me. We bundle prettier with this extension (in which case we know its version) or we search for a prettier along the path (something we don't know before trying to format)

I'll review in details a little later but it seems we were *mostly* using single-quote

azz:
> I'm not sure in which case we need prettier@master ? To format this extension?

To make use of the TypeScript feature, which is not stable enough in the most recent release (v1.3.1).

> This part seems to be complicated to me. We bundle prettier with this extension (in which case we know its version) or we search for a prettier along the path (something we don't know before trying to format, ie already registered our formater)

If prettier is bundles with the extension, this is a non-issue.

If prettier is resolved from the path, you would use semver on `resolved.version` in `requireLocalPkg()` to make sure it supports TypeScript. (lets say hypothetically that TypeScript support lands in prettier@1.5.0, you would check `semver.gte(resolved.version, "1.5.0")`.) If it doesn't mat you'd bail out of `format()`.

CiGit:
Thanks I get it.
I've seen 3 ways to handle missing typescript support in local package:
1. Check it before registering the formater. VSCode will use default TS formater. *This is weird as **some** project path may have prettier. I don't know how to do that*
2. Throw like you said in format.
3. Fallback to bundled one in format. *Similar to previous one, we still format but we hide user's bad setup* 

My preference goes to 2 and 3. I still don't know which one would be more friendly. (IMHO a little more 2)

src/PrettierEditProvider.ts
```diff
@@ -10,41 +10,49 @@ import {
...
     handle deprecated parser option
     */
     let parser = config.parser;
-    if (!parser) { // unset config
-        parser = config.useFlowParser ? 'flow' : 'babylon';
+    if (!parser) {
+        // unset config
+        parser = config.useFlowParser ? "flow" : "babylon";
+    }
+    if (extname(path).startsWith(".ts")) {
```
CiGit: This won't work on an unsaved file. (Untitled)
Could you use [document.languageId](https://code.visualstudio.com/docs/extensionAPI/vscode-api#TextDocument)

You can change format's signature to access the document
Something like:
```js
format(text: string, document: TextDocument): string
```

src/extension.ts
```diff
@@ -4,34 +4,52 @@ import {
...
+        );
     }
+    return config;
 }
+
 export function activate(context: ExtensionContext) {
     const editProvider = new EditProvider();
-    checkConfig();
+    const config = checkConfig();
+    const languageSelector = config.runOnTypeScript
```
CiGit: Nice to test it here.

CiGit:
I don't care personally, but when I started working on this extension it was using single-quote.
@esbenp Is that something you want?
Also 4 spaces indent, semi for what I can remember.

CiGit:
What's happening with eslintIntegration and runOnTypeScript ?

azz:
Made the `languageId` change and reformatted with:
```
"prettier.tabWidth": 4,
"prettier.singleQuote": true,
"prettier.trailingComma": "es5",
"prettier.runOnTypeScript": true,
```

azz:
> What's happening with eslintIntegration and runOnTypeScript ?

One needs to take precedence. At the moment that is eslint.

Behind the scenes, prettier using `typescript-eslint-parser` for TypeScript parsing, so it should be possible to get `prettier-eslint` running on TypeScript code too, but I haven't looked in to that at all yet.

The other alternative would be to switch from `prettier-eslint` to tslint.

CiGit:
Thank you a lot for your work !

Will need to check what happens on typescript files with `prettier-eslint` and adjust the behavior accordingly.

azz:
* prettier/prettier#1835 - Prettier will switch parser to TypeScript if filepath is set. 
* Prettier supports TypeScript from `1.4.0-beta`.

I'll up update the PR shortly.

CiGit:
Great! Can we disable eslintIntegration on ts files? While you are at it :-)

More precisely enable eslintIntegration on `javascript` and `javascriptreact` only. I see CSS also coming (More language *soon*)

CiGit:
Warn : files not yet persisted have no path (Untitled)

azz:
Ended up keeping the language mapping rather than relying on prettier (due to the unsaved file issue).

* Added CSS/LESS/SCSS support.
* Marked prettier dependency as 1.4.0, which will be released soon.

package.json
```diff
@@ -84,10 +89,12 @@
           "type": "string",
           "enum": [
             "babylon",
-            "flow"
+            "flow",
+            "typescript",
+            "postcss"
```
CiGit: I think we should only suggest babylon/flow for JS.
The other one are inferred from language type ?
Or am I missing something?

src/PrettierEditProvider.ts
```diff
@@ -57,16 +77,24 @@ function format(text: string, path: string): string {
         semi: config.semi,
         useTabs: config.useTabs,
     };
+
     if (config.eslintIntegration) {
+        if (isNonJsParser) {
```
CiGit: We should still format with prettier only for those none JS files.
```js
if (config.eslintIntegration && !isNonJsParser )
```

src/PrettierEditProvider.ts
```diff
@@ -57,16 +77,24 @@ function format(text: string, path: string): string {
...
+            filePath: fileName,
+            fallbackPrettierOptions: prettierOptions,
         });
     }
-
-    const prettier = requireLocalPkg(path, 'prettier') as Prettier;
+    const prettier = requireLocalPkg(fileName, 'prettier') as Prettier;
+    if (isNonJsParser && semver.lt(prettier.version, '1.4.0-beta')) {
+        // not supported yet
+        return text;
```
CiGit: I'm not sure silently doing *nothing* is the best option here.
I would throw or silently format with embed version.
Don't know which of these would be better.

Or maybe format with embed and show a warning.

src/types.d.ts
```diff
@@ -1,5 +1,10 @@
-type ParserOption = 'babylon' | 'flow';
-type TrailingCommaOption = 'none' | 'es5' | 'all' | boolean; /* deprecated boolean*/
+type ParserOption = 'babylon' | 'flow' | 'postcss' | 'typescript';
```
CiGit: Same as package.json:
postcss, typescript is not a real choice for the user.

package.json
```diff
@@ -84,10 +89,12 @@
...
           ],
           "default": "babylon",
-          "description": "Which parser to use. Valid options are 'flow' and 'babylon'"
+          "description": "Which parser to use. \"typescript\" and \"postcss\" are supported in prettier 1.4.0"
```
CiGit: May update this one depending on preceding question.

package.json
azz: They are all valid values.

By default, JavaScript files use `babylon`, TypeScript files use `typescript`, and CSS/LESS/SASS files use `postcss`. However you could set the parser to `typescript` on JavaScript files if you really wanted to.

CiGit: Agreed on typescript. But `postcss` throws `SynthaxError`

src/types.d.ts
azz: Not for the user but they're perfectly valid programatically.

package.json
azz: I've changed it back to flow and babylon and changed the description. I feel this option is super misleading as it should only really be set if there's a blocker bug in one parser that forces you to need another.

src/types.d.ts
CiGit: Right 👍

azz:
* Rebased
* Bumped to 1.4.1, 
* changed docs (keep babylon/flow)
* provided a warning when falling back to bundled prettier
* skip eslint if running on non-JS.

CiGit:
Thank you very much ! That's perfect.

We can still add typescript as a parser later. (It may fail on flow tho)

rafaelkallis:
Thanks guys! When can we expect a new version of `prettier-vscode` including `prettier@1.4.1` ?

CiGit:
now :D

PR #112, opened 2017-06-02
Update Prettier to v1.4.0

tilomitra:
Prettier v1.4.0 brings CSS support so it will be good to upgrade to that ASAP. [See Release Notes](https://github.com/prettier/prettier/releases/tag/1.4.0).

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "@types/mocha": "^2.2.32"
   },
   "dependencies": {
-    "prettier": "1.3.1",
+    "prettier": "1.4.0",
     "prettier-eslint": "6.2.3",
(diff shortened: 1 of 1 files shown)


gs-akhan:
Nice !!.  I am jealous, I was hoping I could do it.. 
Any ways also update 

https://github.com/esbenp/prettier-vscode/blob/master/src/extension.ts#L12

by adding "css" and "typescript" prettier 1.4.0 has support for that as well..

@CiGit  Am I right ?
Thanks

CiGit:
There is already #100 which updates prettier and more. Therefor I will close that one. Sorry.
@gs-akhan look at #100 for a complete example ;-)

PR #144, opened 2017-07-01
Update prettier to latest version of package

michaelryancaputo:
There are some issues with the `trailingComma` setting that is causing me to have a bad time. Was hoping this would speed along the update to the latest fixes/changes since 1.5.
Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -218,7 +218,7 @@
     "vscode": "^1.1.0"
   },
   "dependencies": {
-    "prettier": "1.5.0",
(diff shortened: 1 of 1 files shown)


CiGit:
This won't fix the trailing comma issue. 
The fix on prettier's side is not yet published.
However #143 will correct the issue on our side. 
We still want to update the dependency 👍

michaelryancaputo:
Rad, thanks! 👌🏻

Issue #41, opened 2017-02-22
Migrate repository to new prettier organization?


ConAntonakos:
@esbenp Would you be perhaps open to this?

Here is the original Twitter thread: https://twitter.com/ConAntonakos/status/834084544655527937

esbenp:
Yes of course, I think that would be for the better. People like @CiGit have also been working far harder on this than I do so it makes total sense :-)

ConAntonakos:
Fantastic! I created a new issue under the prettier repository: https://github.com/prettier/prettier/issues/781.

ConAntonakos:
Hey, all. Are we still interested in doing this? If not, no worries. I can close this. Pinging @esbenp. Thanks!

RobinMalfait:
It would be nice to do it, just to make it more "official" to new users. But `prettier-vscode` is mentioned as an official plugin on the prettier website. https://prettier.io/

esbenp:
Hey all. I am interested in moving this to the organization since I have no time for working on this. Big shoutout to @CiGit and the rest of you for keeping things moving. The only thing I am not sure on how to do is handing over the VS Code Extension? Does anyone know where I can hand it over to someone else?

CiGit:
It seems the only way to do it (*now*) would be to unpublish and publish with a new publisher. However, there are no redirects, stats etc... This is discouraged for a popular extension. 
> 115,848 installs

I think we belong to this category.

Sources in case I missed something: 
- https://github.com/Microsoft/vscode/issues/3670
- https://github.com/Microsoft/vscode/issues/4772
- https://github.com/Microsoft/vscode/issues/10551

esbenp:
Alright, I do not remember; I added you as a member in the extension admin, right? I do not mind keeping it in my account I just dont have the time to contribute too it. But if you have full access that is good enough for me- then I just need to migrate this repo to the prettier organization. What do you think?

CiGit:
I can publish this extension with your publisher (esbenp). I don't think I can do more.
To handle admin tasks, it would be better if we are 2 at least. OSS is time costly for everyone, I have unavailability too.

azz:
Extension ownership and repository ownership are independent issues, right?

@esbenp You should have sufficient permissions to move this repository to the `prettier` organisation. Do you still want to do that? 

https://help.github.com/articles/transferring-a-repository-owned-by-your-personal-account/#transferring-to-an-organization

esbenp:
[...]

PR #152, opened 2017-07-05
Improve error handling using statusbar item & output channel

RobinMalfait:
Hello!

I was the author of another [prettier extension](https://github.com/RobinMalfait/prettier-eslint-code) which I deprecated for this one.

In the package I wrote I had some nice indications if your code was prettified correctly or not.
 & . Now I know that not everybody wants it so I added an option to enable/disable this feature.

Another thing is, the current error handling is quite annoying in my opinion, you get a popup message with an error and you can't really save the file. For example when you are editing the `settings.json` file from VSCode, you will get errors because it contains invalid JSON, the `// comments`. What this PR enables is that we execute the formatting in a save way and if an error occurs we return the defaultValue (which is the current text), this way we *can* actually edit files with invalid JSON such as `settings.json`.

Last but not least, we still want to view our errors. For this very reason I added an output channel where I write the errors when something happens. If we press the statusbar item we also open the output channel.

Errors look like this:

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -12,6 +12,8 @@ import {
     Position,
 } from 'vscode';
 
+import { safeExecution, addToOutput } from './errorHandler';
+import { onWorkspaceRootChange } from './utils';
 import { requireLocalPkg } from './requirePkg';
 import * as semver from 'semver';
 
@@ -23,7 +25,8 @@ import {
     ParserOption,
 } from './types.d';
 
-type ShowAction = 'Show';
+let errorShown: Boolean = false;
+
 /**
  * Various parser appearance
  */
@@ -35,6 +38,14 @@ const PARSER_SINCE = {
     json: '1.5.0',
     graphql: '1.5.0',
 };
+
+/**
+ * Mark the error as not show, when changing workspaces
+ */
+onWorkspaceRootChange(() => {
+    errorShown = false;
+});
+
 /**
  * Check if the given parser exists in a prettier module.
  * @param parser parser to test
@@ -44,6 +55,7 @@ const PARSER_SINCE = {
 function parserExists(parser: ParserOption, prettier: Prettier) {
     return semver.gte(prettier.version, PARSER_SINCE[parser]);
 }
+
 /**
  * Format the given text with user's configuration.
  * @param text Text to format
@@ -58,6 +70,7 @@ function format(
     const config: PrettierVSCodeConfig = workspace.getConfiguration(
         'prettier'
     ) as any;
+
     /*
     handle trailingComma changes boolean -> string
     */
@@ -110,23 +123,47 @@ function format(
     );
 
     if (config.eslintIntegration && !isNonJsParser) {
-        const prettierEslint = require('prettier-eslint') as PrettierEslintFormat;
-        return prettierEslint({
+        return safeExecution(
+            () => {
+                const prettierEslint = require('prettier-eslint') as PrettierEslintFormat;
+                return prettierEslint({
+                    text,
+                    filePath: fileName,
+                    fallbackPrettierOptions: prettierOptions,
+                });
+            },
             text,
-            filePath: fileName,
-            fallbackPrettierOptions: prettierOptions,
-        });
+            fileName
+        );
     }
     const prettier = requireLocalPkg(fileName, 'prettier') as Prettier;
     if (isNonJsParser && !parserExists(parser, prettier)) {
-        const bundledPrettier = require('prettier') as Prettier;
-        window.showWarningMessage(
-            `prettier@${prettier.version} doesn't support ${languageId}. ` +
-                `Falling back to bundled prettier@${bundledPrettier.version}.`
+        return safeExecution(
+            () => {
+                const bundledPrettier = require('prettier') as Prettier;
+                const warningMessage =
+                    `prettier@${prettier.version} doesn't support ${languageId}. ` +
+                    `Falling back to bundled prettier@${bundledPrettier.version}.`;
+
+                addToOutput(warningMessage);
+
+                if (errorShown === false) {
+                    window.showWarningMessage(warningMessage);
+                    errorShown = true;
+                }
+
+                return bundledPrettier.format(text, prettierOptions);
+            },
+            text,
+            fileName
         );
-        return bundledPrettier.format(text, prettierOptions);
     }
-    return prettier.format(text, prettierOptions);
+
+    return safeExecution(
+        () => prettier.format(text, prettierOptions),
+        text,
+        fileName
+    );
 }
 
 function fullDocumentRange(document: TextDocument): Range {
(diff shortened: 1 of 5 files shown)


CiGit:
Thanks. I've no time to test it right now but definitely will 😄 
I will asked it here:
I love prettier, I hate options. Is that option really necessary? IMO the status bar indicator could be even smaller : 
> P: ❌     P: ✔️    with a tooltip

Is it possible to insert clickable links in those outputs? In your screenshot I'd like to click on the error and being brought to the location. Like `Show` currently.
If not, I would add the file which produced the error in the output text.

RobinMalfait:
@CiGit I hate options to, I will delete the option for now, and if people request it we can re-enable it if necessary.
I don't really like the shorter version of `P: ...` because if you have more extensions that do that it can get messy in that bar. `Prettier: ...` is short enough in my opinion.

DemianD:
I would also use "Prettier" instead of just "P" because I use a couple other plugins starting with a P like Path Intellisense, PHP Debug, PHP Intellisense, ...

But it's just a preference
(Make it cucustomizable with an option 🤣)

CiGit:
BTW: 
> we execute the formatting in a save way and if an error occurs we return the defaultValue (which is the current text), this way we can actually edit files with invalid JSON such as settings.json

It was already possible. Failing on format still saves it, with a ugly error message ;-)

---
I've played with it a little.
Now we do not see errors anymore. That's good. Note: I've searched for `Prettier:...` on the right side first.
We see only the last `error/all good` in the output. Meaning if I format multiple files, I get only the last output, silencing the rest.
`outputChannel.clear();`.
What I had in mind was using it like an append only console. The user can still clear it.

RobinMalfait:
@CiGit oh cool, let me update the output a little bit.

RobinMalfait:
In the last commit, I made it possible to append to the output channel. It was quite a mess so I added some *titles*, what do you think?

CiGit:
I like it.
Maybe don't output "All good" that's just noise.

RobinMalfait:
Done! :D

CiGit:
And maybe remove the '\n' between `----------` and the error. This will group timestamp with the corresponding message.

I'll review the code a little latter. In the meantime can you rebase and squash all those commits, except the one which removes the option (if we want to revert once)

RobinMalfait:
Sure, I'll squash them together.

src/PrettierEditProvider.ts
```diff
@@ -44,6 +47,61 @@ const PARSER_SINCE = {
 function parserExists(parser: ParserOption, prettier: Prettier) {
     return semver.gte(prettier.version, PARSER_SINCE[parser]);
 }
+
+let statusbar: StatusBarItem = undefined;
```
CiGit: We should create it directly, not on first run. It gives the information: Prettier is up !

src/PrettierEditProvider.ts
```diff
@@ -110,23 +169,30 @@ function format(
...
     if (isNonJsParser && !parserExists(parser, prettier)) {
-        const bundledPrettier = require('prettier') as Prettier;
-        window.showWarningMessage(
-            `prettier@${prettier.version} doesn't support ${languageId}. ` +
-                `Falling back to bundled prettier@${bundledPrettier.version}.`
-        );
-        return bundledPrettier.format(text, prettierOptions);
+        return safeExecution(() => {
+            const bundledPrettier = require('prettier') as Prettier;
+            window.showWarningMessage(
```
CiGit: I would also put this message into the output.

src/PrettierEditProvider.ts
```diff
@@ -110,23 +169,30 @@ function format(
...
+                `prettier@${prettier.version} doesn't support ${languageId}. ` +
+                    `Falling back to bundled prettier@${bundledPrettier.version}.`
+            );
+            return bundledPrettier.format(text, prettierOptions);
+        }, text, fileName);
     }
-    return prettier.format(text, prettierOptions);
+
+    return safeExecution(() => {
+        return prettier.format(text, prettierOptions);
```
CiGit: Could be a simple arrow (Non-block)

CiGit:
We now have the same problem with the status icon, it shows only the last run ( ❌ ✔️ )
.

RobinMalfait:
Regarding to your last comment

> We now have the same problem with the status icon, it shows only the last run ( ❌ ✔️ )

How would you handle this? For error output it is easy, we can just append the the outputChannel. For this, I don't want to do management of files I guess? Or would you display the information based on the current active file? When you switch tabs to another file, display the correct information there?

CiGit:
Yeah this is something to think about...
It feels just weird to have ✔️ with errors.
I've some ideas:
 - show the ❌ once an error occurred. Remove it on clear if possible.
 - Have an error marker on each file. And show it depending on the focus. Something harder to do.
 - Show nothing...

This could be an other PR tho. I find this one is a good enhancement.
Let me know your thought. 
🛌

src/PrettierEditProvider.ts
RobinMalfait: Applied in cc15e68

src/PrettierEditProvider.ts
RobinMalfait: Applied in cc15e68

src/PrettierEditProvider.ts
RobinMalfait: Applied in cc15e68

RobinMalfait:
I also think that it should be a future improvement and in a separate PR. I like this one the most:

> Have an error marker on each file. And show it depending on the focus. Something harder to do.

Because in the first option you have to keep track of all the items. And last option is not ideal because I love the ❌  & ✔️

CiGit:
Seems we misunderstood ourself.
>Because in the first option you have to keep track of all the items.
That's not what I meant.
Error -> :x: -> outputChannel. on("clear") -> :heavy_check_mark: 

In the second option, you will have to keep a list of files which have errors.

New idea:
Don't use status bar item. Open the output channel on first error or provide an action "show prettier errors"

src/errorHandler.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Setup the output channel and the statusBarItem.
+ * Create a command to show the output channel
+ * 
+ * @returns {Disposable} The command to open the output channel
+ */
+export function setupErrorHandler(): Disposable {
+    // Setup the statusBarItem
+    statusBarItem = window.createStatusBarItem();
+    statusBarItem.text = 'Prettier is up!';
```
CiGit: Didn't meant to have that message :smile: 
Just 'Prettier'

src/PrettierEditProvider.ts
```diff
@@ -110,23 +113,41 @@ function format(
...
-        window.showWarningMessage(
-            `prettier@${prettier.version} doesn't support ${languageId}. ` +
-                `Falling back to bundled prettier@${bundledPrettier.version}.`
+        return safeExecution(
+            () => {
+                const bundledPrettier = require('prettier') as Prettier;
+                const warningMessage =
+                    `prettier@${prettier.version} doesn't support ${languageId}. ` +
+                    `Falling back to bundled prettier@${bundledPrettier.version}.`;
+                window.showWarningMessage(warningMessage);
```
CiGit: I would remove that popup completely. It may occur a lot. Or better : show it only once.

RobinMalfait:
I think that opening the output channel on error will be very annoying, especially when you have the `editor.formatOnType` set to `true`.

CiGit:
It will only open the first time ie on create

RobinMalfait:
Okay another proposal:

What if we do a `dry_run` everytime we switch to an open file. If an error occurs show `Prettier: X` else `Prettier: V` but don't really save it?

> It will only open the first time ie on create

That's also an option. Which do you prefer?

src/errorHandler.ts
RobinMalfait: Haha ok hold on ^^

src/PrettierEditProvider.ts
RobinMalfait: Only once is quite hard no? If you open another workspace or something it may be weird behavior.

CiGit:
I don't think your proposal fits prettier's role. IMO that's a linter's role.

src/PrettierEditProvider.ts
CiGit: You are right.
Dummy implementation:
let a boolean 'errorShown' = false
don't show that error if errroShown
set it to true once shown.
set it to false on workspace change event. (I pretty sure that one exists)

RobinMalfait:
Last changes include:

- Show warning once per workspace
- Open output channel on first error

CiGit:
Seems to me:
Open output channel on first error *per workspace*

I could nit-pick, but LGTM.
We can still change things later on. Great job on this first iteration :-)

Will let it open for some *potential* reviews.
Will merge it this evening and release.
Please squash :-)

RobinMalfait:
I'll squash everything together. If we ever need to re-add the option to show/hide the statusbar it's a small change.

CiGit:
Thanks a lot for your work. It has been released! 
If you agree, we should ask @esbenp to give you the same status I have. You seemed involved to me.

RobinMalfait:
@CiGit thank you! I love prettier and loved building my extension. But there is no need to maintain multiple extensions. 1 good extension is key.

With status you mean collaborator and be allowed to merge PR's? I'll try my best to help wherever I can to improve the package even more :)

CiGit:
Yes  `Collaborator`.
This means having the right to merge PR, push to master, close issues ........... and maybe even publish if we can have this from @esbenp (who vanished).
I feel so lonely here :-)

RobinMalfait:
Hehe, it might also be an option to move the prettier-vscode repo to the prettier namespace? That way we don't have to spam @esbenp so much :p and/or there will be more issues & PR's?

CiGit:
He is the one who can move that repo. #41 
There are always issues, PR. There is 1 PR awaiting. And there is more to do here.

PR #156, opened 2017-07-06
add support for prettier with semistandard and standard

bySabi:
Add support for `prettier + semistandard` and  `prettier + standard`

Tested locally with create VSIX and works 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
+## [0.21.0]
+- Reworked error messages. They are now in a dedicated output channel.
+
 ## [0.20.0]
 - Prettier 1.5
 - Added JSON and GraphQL formatting (Range formatting disabled)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,3 +3,4 @@
 - [David Burrows](https://github.com/dburrows)
 - [Jon Wolfe](https://github.com/JonathanWolfe)
 - [Laurence Rowe](https://github.com/lrowe)
+- [bySabi Files](https://github.com/bySabi)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,3 +4,4 @@
 - [Jon Wolfe](https://github.com/JonathanWolfe)
 - [Laurence Rowe](https://github.com/lrowe)
 - [Robin Malfait](https://github.com/RobinMalfait)
+- [bySabi Files](https://github.com/bySabi)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,4 +3,4 @@
 - [David Burrows](https://github.com/dburrows)
 - [Jon Wolfe](https://github.com/JonathanWolfe)
 - [Laurence Rowe](https://github.com/lrowe)
-- [bySabi Files](https://github.com/bySabi)
+- [Robin Malfait](https://github.com/RobinMalfait)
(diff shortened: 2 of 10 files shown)


bySabi:
@CiGit please review this PR

CiGit:
I won't review this before I know what I should expect.
What's the added value ?
What's the outcome ?

At first sight I see a (semi)standard formater. Here is prettier.....
Sorry if I feel rude. But this seems like a ninja PR to me.

bySabi:
The proposal is about and option to switch `prettier-eslint` engine when `prettier.eslintIntegration` is enable.

I use `semistandard/standard` in my daily `lint && test` workflow but their linter rules conflict with default `prettier-eslint` rules. I know it can be customized trought `.eslint` files but I like **zero conf** tools like `semistandard/standard`.

That´s why I create some modules for integrate `prettier`  and `semistandard/standard`
* https://github.com/bySabi/prettier-semi
* https://github.com/bySabi/prettier-std
* https://github.com/bySabi/prettier-semi-cli
* https://github.com/bySabi/prettier-std-cli

You can see in the doc of [prettier-eslint-cli](https://github.com/prettier/prettier-eslint-cli#integration) some info about this tools.

`prettier-eslint` have an option for inject the `eslint` module: https://github.com/prettier/prettier-eslint/blob/master/src/index.js#L24 but is not posible inject a **not already installed module** on a Visual Studio extension. That's why this PR include `prettier-semi`  and `prettier-std` packages
and the **magic** options `type PrettierEslintOption = 'prettier-eslint' | 'prettier-semi' | 'prettier-std';`

bySabi:
Maybe we could tweak `prettier.eslintIntegration` a little, say:

`prettier.eslintIntegration: [boolean]`  for backward compatibility and a new sintax, Ex:

```
"prettier.eslintIntegration": {
  "eslintEngine": "standard",
  "enable": true
}
```
* `enable`  will be the same has `prettier.eslintIntegration: true`
* `eslintEngine` will be `eslint | standard | semistandard`  (default: 'eslint')
```

What you think?

bySabi:
Well it things that VSCode developers recommed single key/value pairs for configuration https://github.com/Microsoft/vscode/issues/20791

`prettier.eslintEngine: ( 'eslint' | 'standard' | 'semistandard' )`  new option looks good.

bySabi:
Update PR with latest commit to `esbenp/prettier-vscode` master.

CiGit:
If I get it right, this is a (semi)standard formater.
I don't want to support it in this extension. There is an extension for standard. And if you really want, you can still do it with eslint + eslint-config-standard.
The purpose is not to support every formater/config out there, managing plenty of deps and options.

bySabi:
> If I get it right, this is a (semi)standard formater. 

This is not formatters: https://www.npmjs.com/package/prettier-standard or  https://www.npmjs.com/package/prettier-semistandard

Is just `prettier-eslint` formatter with the option `--eslintPath` set.  The code is soo simple that can be integrate in the extension: https://github.com/bySabi/prettier-std/blob/master/index.js

> I don't want to support it in this extension. There is an extension for standard. And if you really want, you can still do it with eslint + eslint-config-standard.

I don´t see how I can use `prettier-eslint` with `eslint-config-standard`. **standard** need all this modules: eslint-config-standard eslint-plugin-standard eslint-plugin-promise eslint-plugin-import eslint-plugin-node

> The purpose is not to support every formater/config out there, managing plenty of deps and options.

You are right here. But I cannot figure out how add  externals packages to this extension, maybe with plugin approach?

`standard` is use widely https://github.com/standard/standard#who-uses-javascript-standard-style. Include  a explicit support will increase the popularity of this extension

bySabi:
@CiGit do you know how we can implement a plugin system with the extension?

CiGit:
>This is not formatters

vs
>Is just prettier-eslint formatter 

:confused: 

Correction on my previous assertion: 
You can already do that with : `prettier-eslint` + 
> eslint-config-standard eslint-plugin-standard eslint-plugin-promise eslint-plugin-import eslint-plugin-node

I don't know how plugin would work. Never investigate that direction. May be worth tho.
What I know is that extensions may depend on other extensions. https://code.visualstudio.com/docs/extensionAPI/extension-manifest

And if you really want to have a _no config_ tooling, you can fork this extension remove all options and make the necessary tweaks to support `prettier-(semi)standard`
You may first have a look at https://marketplace.visualstudio.com/items?itemName=numso.prettier-standard-vscode which is a fork of this one and seems to do what you want.

PR #166, opened 2017-07-13
Override the package name which is used to search in local dependencies

Alphapage:
Related to https://github.com/esbenp/prettier-vscode/issues/164

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,9 @@ Advanced feature. Use this to opt in / out prettier on various language ids. Res
 Use parser `postcss` for given language ids.
 Use with care.
 
+#### prettier.packageName (default: 'prettier') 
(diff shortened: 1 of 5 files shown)


CiGit:
I'm not that convinced we want such an option. I think this will cause more troubles than bringing solutions.

Alphapage:
No problem. It was just a helper for my particular problem.

CiGit:
You may want to maintain a fork for your *particular problem*.
This seems to create a conflict between eslintIntegration and this option ;-) prettier-eslint will still search for a *prettier* package.

Alphapage:
Yes, this will probably cause problems for prettier dependent packages, even if prettier-eslint has an option to set prettierPath.

bySabi:
@Alphapage I wrote a script that fork `prettier-vscode`, patch, create VSIX and publish.
Maybe can be useful to you:

https://github.com/bySabi/patch--prettier-vscode

Alphapage:
Interesting. I will investigate this option. Thank you.

PR #167, opened 2017-07-15
File system configuration files

### LINKED ISSUE #124: Can the config be shared with command line use of prettier.
For example it makes sense to have prettier set to run on save but the exact same settings will be needed for a git hook or CI processing

As far as I can see as Prettier doesn't use package.json for settings but the command line this won't be possible without an upstream patch.

### LINKED ISSUE #130: Support different trailingComma settings per language
Our project uses `trailingComma: all` for typescript and `trailingComma: es5` for javascript.  Is there a way to set that up here?

### LINKED ISSUE #163: Support Prettier config from prettierrc and package.json
Recently, Prettier added support for storing the configuration in a project file.

https://github.com/prettier/prettier#configuration-file
> Prettier uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) for configuration file support. This means you can configure prettier via:
>
> - A .`prettierrc` file, written in YAML or JSON.
> - A `prettier.config.js` file that exports an object.
> - A `"prettier"` key in your `package.json` file.
> 
> The configuration file will be resolved starting from the location of the file being formatted, and searching up the file tree until a config file is (or isn't) found.

AFAICT, the VSCode extension ignores this config now.

It would be great if the extension could load the config the same way how Prettier CLI does it, and then applying any VSCode-specific configuration as an override on top of that base configuration.

CiGit:
Brings support for reading configuration files:
 - .prettierrc
 - package.json
 - prettier.config.js

Also validates ([JSON Schema](http://json.schemastore.org/prettierrc)) .prettierrc files.

This depends on @azz 's work prettier/prettier#2434 and prettier/prettier#2451 being released.
ie: https://github.com/prettier/prettier/commit/3136907a1556391f88e72d1935d2e8f1ae4e8d33 must have a tag.

fixes #163 
fixes #124 
partial fix #130 

Todo:
- [x] Readme
- [x] Based on a released prettier

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,27 +26,23 @@ OR
 ```
 
 ### Format On Save
-
 Respects `editor.formatOnSave` setting.
 
 ## Settings
 
 ### Prettier's Settings
-
 Settings will be read from:
 1. File system, first matching file in 
     1. `package.json` - `prettier` key
     1. `.prettierrc`
     1. `.prettier.config.js`
-1. VSCode prettier's settings, described bellow
+1. VSCode prettier's settings, described below
 1. VSCode prettier's default settings
 
 #### prettier.printWidth (default: 80)
-
 Fit code within this line limit
 
 #### prettier.tabWidth (default: 2)
-
 Number of spaces it should use per tab
 
 #### prettier.singleQuote (default: false)
@@ -99,5 +95,4 @@ Use with care.
 This extension will use prettier from your project's local dependencies. Should prettier not be installed locally with your project's dependencies, a copy will be bundled with the extension. 
 
 ## Contribute
-
-This is my first Visual Studio Extension so I probably made some terrible choices. Feel free to open issue or PRs!
+This is my first Visual Studio Extension so I probably made some terrible choices. Feel free to open issues or PRs!
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 VS Code package to format your Javascript / Typescript / CSS using [Prettier](https://github.com/prettier/prettier).
 
-### Installation
+## Installation
 
 Install through VS Code extensions. Search for `Prettier - JavaScript formatter`
 
@@ -14,9 +14,9 @@ Can also be installed using
 ext install prettier-vscode
 ```
 
-### Usage
+## Usage
 
-#### Using Command Palette (CMD + Shift + P)
+### Using Command Palette (CMD/CTRL + Shift + P)
 
 ```
 1. CMD + Shift + P -> Format Document
@@ -25,14 +25,21 @@ OR
 2. CMD + Shift + P -> Format Selection
 ```
 
-#### Format On Save
+### Format On Save
 
 Respects `editor.formatOnSave` setting.
 
-### Settings
-#### prettier.eslintIntegration (default: false) - Javascript only
-Use *[prettier-eslint](https://github.com/prettier/prettier-eslint)* instead of *prettier*.
-Other settings will only be fallbacks in case they could not be inferred from eslint rules.
+## Settings
+
+### Prettier's Settings
+
+Settings will be read from:
+1. File system, first matching file in 
+    1. `package.json` - `prettier` key
+    1. `.prettierrc`
+    1. `.prettier.config.js`
+1. VSCode prettier's settings, described bellow
+1. VSCode prettier's default settings
 
 #### prettier.printWidth (default: 80)
 
@@ -67,6 +74,12 @@ or only at the beginning of lines that may introduce ASI failures (semi: false)
 #### prettier.useTabs (default: false)
 If true, indent lines with tabs
 
+### VSCode specific settings
+
+#### prettier.eslintIntegration (default: false) - Javascript only
+Use *[prettier-eslint](https://github.com/prettier/prettier-eslint)* instead of *prettier*.
+Other settings will only be fallbacks in case they could not be inferred from eslint rules.
+
 #### prettier.javascriptEnable (default: ["javascript", "javascriptreact"])
 Advanced feature. Use this to opt in / out prettier on various language ids. Restart required.
 Use parser `babylon` or `flow` depending on `prettier.parser` for given language ids.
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -219,7 +233,7 @@
     "vscode": "^1.1.0"
   },
   "dependencies": {
-    "prettier": "1.5.2",
+    "prettier": "github:prettier/prettier",
```
CiGit: Must be a released version

azz:
Awesome!

CiGit:
ClearConfigCache does it's job.

Explanation on settings priority:
 1. First encountered file config. File path upwards
 2. VSCode's prettier settings
 3. Prettier's default settings (In fact VSCode's prettier default settings)

@azz do you know what happens when there are multiple files at same path level? ie
```
. 
 |- .prettierrc
 |- package.json "prettier"
 |- prettier.config.js
```

azz:
Yes, the order is:

1. package.json
2. .prettierrc
3. prettier.config.js

CiGit:
Thank you!

package.json
```diff
@@ -202,7 +202,21 @@
...
-    }
+    },
+    "jsonValidation": [
+      {
+        "fileMatch": ".prettierrc",
+        "url": "http://json.schemastore.org/prettierrc"
+      }
+    ],
+    "languages": [
+      {
```
azz: We also support YAML in `.prettierrc` files but it doesn't make sense to have two files in one language mode...

src/configCacheHandler.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Create a file watcher. Clears prettier's configuration cache on
+ * file change, create, delete.
+ * @returns disposable file system watcher.
+ */
+function fileListener() {
+    const fileWatcher = workspace.createFileSystemWatcher(
+        `**/{${PRETTIER_CONFIG_FILES.join(',')}}`,
+        false,
```
azz: From memory, false is the default for these flags.

package.json
CiGit: Yeah I've not found a solution to that.

src/configCacheHandler.ts
CiGit: You are right.

danawoodman:
This would be awesome to have as currently the extension is not usable for projects that use a custom config. Any idea when this will get merged in?

CiGit:
@danawoodman It depends on prettier. Certainly in prettier 1.6 which should be near.

README.md
```diff
@@ -82,9 +95,9 @@ Advanced feature. Use this to opt in / out prettier on various language ids. Res
...
 Use with care.
 
-### Prettier resolution
+## Prettier resolution
 This extension will use prettier from your project's local dependencies. Should prettier not be installed locally with your project's dependencies, a copy will be bundled with the extension. 
 
-### Contribute
+## Contribute
 
 This is my first Visual Studio Extension so I probably made some terrible choices. Feel free to open issue or PRs!
```
glennreyes: open _an_ issue

README.md
```diff
@@ -25,14 +25,21 @@ OR
...
+## Settings
+
+### Prettier's Settings
+
+Settings will be read from:
+1. File system, first matching file in 
+    1. `package.json` - `prettier` key
+    1. `.prettierrc`
+    1. `.prettier.config.js`
+1. VSCode prettier's settings, described bellow
```
glennreyes: below

README.md
RobinMalfait: or _issues_

CiGit: issues.
We should update that part once ... If someone has an idea.

src/PrettierEditProvider.ts
```diff
@@ -22,6 +22,7 @@ import {
     ParserOption,
 } from './types.d';
 
+const bundledPrettier = require('prettier') as Prettier;
```
airato: I'm wondering what's the reason to not use es6 imports here and in src/configCacheHandler.ts? like `import prettier from 'prettier';`

CiGit: I would say laziness with typings now. Before it could be to defer the loading, but that's not the case anymore.

CiGit:
Rebased.

src/PrettierEditProvider.ts
azz: Can you still do

```ts
declare module "prettier" { export = Prettier }
import * as prettier from prettier;
```
?

CiGit: I'm not that comfortable with typescript modules. This should be possible I think.
I would simply have rewritten it.

danawoodman:
So now that Prettier has pushed `v1.6.0-beta` that has support for config files, could we get this merged in? Will it use locally installed Prettier? See [here](https://github.com/prettier/prettier/issues/2589) for note on recent release

CiGit:
Would wait for a final release. And reading config is done from bundled prettier. formatting is done with your locally installed prettier.

RobinMalfait:
It's ready!

danawoodman:
@RobinMalfait what do you mean by it's ready? It looks like the PR is still open?

hyperwidget:
@danawoodman - pretty sure that @RobinMalfait was referring to the fact that Prettier itself released v 1.6 today which includes support for .prettierrc files - which means that @CiGit 's concern about waiting for final release should be alleviated :D

danawoodman:
@hyperwidget 👍 sweet. I assume we still need to wait for this PR to be merged or will it work if we update local Prettier?

azz:
If you want config files to work with the extension you'll need to wait until this is merged and released.

RobinMalfait:
Exactly, I was referring to the 1.6 release of prettier :D

tsiq-swyx:
hello just checking in - adding 
```
  "prettier": {
    "singleQuote": "true"
  },
```

to my package.json in VSCode doesnt work :(

olstenlarck:
Use dot files for everything and you VSCode will love you. I hated long time the dot files, but they are must, that's why all of my recent 7 months of repos in @tunnckoCore are full of dotfiles.

CiGit:
@tsiq-swyx please open a new issue.

tsiq-swyx:
its ok for now, i just used a dotfile and it was fine. i'll report it when i have more time to follow up, i dont right now. thanks for responding

PR #175, opened 2017-07-21
catch module resolution errors

### LINKED ISSUE #174: Extension not falling back to bundled prettier on windows
A colleague at work today asked me for help with this extension. Format wasn't working at all (not on save or Format Document). There were no error messages and the Prettier output panel was not present. This happened on Windows 10 and persisted past reinstalling the extension.

Turned out that prettier wasn't installed locally (running `yarn` fixed it), but to my knowledge the extension should notify and fallback to the bundled version of prettier (1.5.2) when no local package is found.

Either way there is a UX problem. I'd expect an error message upon the format comment being invoked.

Unable to provide detailed reproduction steps unfortunately.

Config was just:

```json
{
  "editor.formatOnSave": true,
  "prettier.tabWidth": 4
}
```

We only tried formatting TypeScript files.

CiGit:
Fallback to bundled package if loading it locally failed. Log to output channel.

Fixes #174

### DIFF
--- a/src/requirePkg.ts
+++ b/src/requirePkg.ts
@@ -1,3 +1,5 @@
+import { addToOutput } from './errorHandler';
+
 const path = require('path');
 const readPkgUp = require('read-pkg-up');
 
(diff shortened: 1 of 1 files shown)


RobinMalfait:
LGTM 👍

Issue #178, opened 2017-07-24
Status bar: Move "Prettier" from left to right


vvs:
I think it would be nice to be consistent with other extensions, and move "Prettier" to the right.

Here's how it is now:
![prettier](https://user-images.githubusercontent.com/9781/28531060-f186913e-7095-11e7-86b0-69eed0755086.png)

RobinMalfait:
What are your thoughts on this one @CiGit ?

CiGit:
I'm still searching it on the right side :-) Certainly still used to eslint being on the right side. [this](https://github.com/esbenp/prettier-vscode/pull/152#issuecomment-313204170) was my first impression.

I think, for consistency, it could be on the right side, even if I think we can get used to it.

RobinMalfait:
@CiGit I'm working on a PR to fix #161 maybe I can include this? Like adding it to the right side?

Issue #161, opened 2017-07-09
Prettier status bar should be hidden for unsupported languages


szastupov:
Hi. I like the new error panel but would prefer it to be hidden when I don't work with JavaScript. Currently it's always shown. Maybe this can be solved by bringing back proper activationEvents to package.json?

CiGit:
Yes we could think about it.
>Maybe this can be solved by bringing back proper activationEvents to package.json?

I'd like to do this, but the activation event depends on all `prettier. ...Enable` settings.

szastupov:
Hm, I'm not very aware of vscode api, but can activationEvents and extension.activate be used at the same time? Like, we could fill activationEvents with all supported languages and then do additional check inside activate().

CiGit:
The problem we have is we don't know "all supported languages". Extensions may register their language.

@RobinMalfait Any thought on this?

szastupov:
Yeah, but you're in sync with latest version of prettier, so at least you know what languages this version of prettier supports. Also, does those advanced options are widely used? Because if they're not, then it doesn't worth supporting them :) It's just better to be a nice citizen and provide proper activationEvents, especially if it's visible that plugin is active.

I'm not pushing, just trying to elaborate. I will keep using the extension anyway 😘

RobinMalfait:
I think that we should not rely on those activation events because once the plugin is activated it's activated. Also the "supported languages" is a problem because it is so configurable. What we should do is toggle the the statusbar item.

Question:

- What do you expect when you are working on javascript file and open a yaml file?
  - Do you want to hide the statusbar because it is not a supported language?
  - Do you want to keep it there because you worked in a javascript right before it?

szastupov:
> Do you want to hide the statusbar because it is not a supported language?

Yep, I think that's the desired behaviour. I.e. show only for supported languages, hide when I switch to something else.

RobinMalfait:
In that case the we can't rely on activation events, so we should watch for workspace document changes and toggle the status bar item when the current language is not in your list of enabled languages. @CiGit Do you want me to provide a PR for this?

CiGit:
I'm happy to let you "statusbar" issues :-) (others too)


>Do you want to hide the statusbar because it is not a supported language?
Do you want to keep it there because you worked in a javascript right before it?

Both seems correct behavior to me.
1. Due to it's action (open panel) it may seems relevant to always have that button if there is something in the panel
2. Explicitly tells : Prettier runs here.

[...]

PR #179, opened 2017-07-24
Only show status bar when the active language is `enabled`

RobinMalfait:
This PR fixes both #161 and #178

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -82,10 +82,6 @@ function format(
     */
     let parser = config.parser;
     let isNonJsParser = false;
-    if (!parser) {
-        // unset config
-        parser = 'babylon';
-    }
     if (config.typescriptEnable.includes(languageId)) {
         parser = 'typescript';
         isNonJsParser = true;
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -84,7 +84,7 @@ function format(
     let isNonJsParser = false;
     if (!parser) {
         // unset config
-        parser = config.useFlowParser ? 'flow' : 'babylon';
+        parser = 'babylon';
     }
     if (config.typescriptEnable.includes(languageId)) {
         parser = 'typescript';
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -122,8 +122,8 @@ export function setupErrorHandler(): Disposable {
     // Setup the statusBarItem
     statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right, -1);
     statusBarItem.text = 'Prettier';
-    statusBarItem.show();
     statusBarItem.command = 'prettier.open-output';
+    statusBarItem.hide();
 
     // Setup the outputChannel
     outputChannel = window.createOutputChannel('Prettier');
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -114,7 +114,7 @@ export function safeExecution(
  */
 export function setupErrorHandler(): Disposable {
     // Setup the statusBarItem
-    statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right);
+    statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right, -1);
     statusBarItem.text = 'Prettier';
     statusBarItem.show();
 
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -17,6 +17,10 @@ let outputChannelOpen: Boolean = false;
 
 function toggleStatusBarItem(editor: TextEditor): void {
     if (editor !== undefined) {
+        // The function will be triggered everytime the active "editor" instance changes
+        // It also triggers when we focus on the output panel or on the debug panel
+        // Both are seen as an "editor".
+        // The following check will ignore such panels
         if (
             ['debug', 'output'].some(
                 part => editor.document.uri.scheme === part
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -14,31 +14,34 @@ let statusBarItem: StatusBarItem;
 let outputChannel: OutputChannel;
 let outputChannelOpen: Boolean = false;
 
-/**
- * Mark the outputChannelOpen as false when changing workspaces
- */
-onWorkspaceRootChange(() => {
-    outputChannelOpen = false;
-});
+export function registerDisposables(): Disposable[] {
+    return [
+        // Mark the outputChannelOpen as false when changing workspaces
+        onWorkspaceRootChange(() => {
+            outputChannelOpen = false;
+        }),
 
-window.onDidChangeActiveTextEditor(editor => {
-    if (statusBarItem !== undefined) {
-        if (editor !== undefined) {
-            const score = languages.match(
-                allEnabledLanguages(),
-                editor.document
-            );
+        // Keep track whether to show/hide the statusbar
+        window.onDidChangeActiveTextEditor(editor => {
+            if (statusBarItem !== undefined) {
+                if (editor !== undefined) {
+                    const score = languages.match(
+                        allEnabledLanguages(),
+                        editor.document
+                    );
 
-            if (score > 0) {
-                statusBarItem.show();
-            } else {
-                statusBarItem.hide();
+                    if (score > 0) {
+                        statusBarItem.show();
+                    } else {
+                        statusBarItem.hide();
+                    }
+                } else {
+                    statusBarItem.hide();
+                }
             }
-        } else {
-            statusBarItem.hide();
-        }
-    }
-});
+        }),
+    ];
+}
 
 /**
  * Update the statusBarItem message and show the statusBarItem
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -17,6 +17,14 @@ let outputChannelOpen: Boolean = false;
 
 function toggleStatusBarItem(editor: TextEditor): void {
     if (editor !== undefined) {
+        if (
+            ['debug', 'output'].some(
+                part => editor.document.uri.scheme === part
+            )
+        ) {
+            return;
+        }
+
         const score = languages.match(allEnabledLanguages(), editor.document);
 
         if (score > 0) {
@@ -24,8 +32,6 @@ function toggleStatusBarItem(editor: TextEditor): void {
         } else {
             statusBarItem.hide();
         }
-    } else {
-        statusBarItem.hide();
     }
 }
 
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -2,11 +2,13 @@ import {
     Disposable,
     StatusBarItem,
     OutputChannel,
+    StatusBarAlignment,
     commands,
     window,
+    languages,
 } from 'vscode';
 
-import { onWorkspaceRootChange } from './utils';
+import { onWorkspaceRootChange, allEnabledLanguages } from './utils';
 
 let statusBarItem: StatusBarItem;
 let outputChannel: OutputChannel;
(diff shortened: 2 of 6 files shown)


RobinMalfait:
The `score` thing might be a bit weird, but it comes straight from the docs:

RobinMalfait:
@szastupov @vvs You both added an issue, this PR fixes it. Some review from both of you would be nice :)

szastupov:
Looks good 👍

vvs:
@RobinMalfait Would you mind telling me how to install the updated extension from the repository?

I cloned the repo, opened the folder in VS Code, and then what?

src/errorHandler.ts
```diff
@@ -100,7 +114,7 @@ export function safeExecution(
  */
 export function setupErrorHandler(): Disposable {
     // Setup the statusBarItem
-    statusBarItem = window.createStatusBarItem();
+    statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right);
```
vvs: You need to provide the second argument (priority) with non-zero value to avoid intermittent "jumping" of "Prettier" position back and forth. It looks like ESLint label has priority 0, so we need to set priority for Prettier to 1 or -1.

-1 would permanently move the "Prettier" label to the top-right corner.

vvs:
Here's what I see without changing the priority:
vs_code-prettier

src/errorHandler.ts
RobinMalfait: Gotcha let me change that quickly ^^

RobinMalfait:
@vvs did you figure it out on how to run it locally or not?
Basically an npm install and press F5 to run it. A new vscode version will popup with the plugin loaded.

vvs:
@RobinMalfait I managed to run the updated extension via command line: `code.cmd --extensionDevelopmentPath=W:\work\github\prettier-vscode.git`

It doesn't work for me via F5 for some reason:
```
Exception has occurred: Error
Error: Cannot find module 'vscode'
```

Back to PR though, everything else looks good and works, thanks!

RobinMalfait:
OH okay, you have to run `npm install`, if you are a yarn user that won't work. (at least not for me)

src/errorHandler.ts
```diff
@@ -19,6 +21,18 @@ onWorkspaceRootChange(() => {
     outputChannelOpen = false;
 });
 
+window.onDidChangeActiveTextEditor(editor => {
+    if (editor !== undefined && statusBarItem !== undefined) {
```
CiGit: I think we should hide it when there is no editor.

src/errorHandler.ts
```diff
@@ -100,7 +114,7 @@ export function safeExecution(
  */
 export function setupErrorHandler(): Disposable {
     // Setup the statusBarItem
-    statusBarItem = window.createStatusBarItem();
+    statusBarItem = window.createStatusBarItem(StatusBarAlignment.Right, -1);
     statusBarItem.text = 'Prettier';
     statusBarItem.show();
```
CiGit: We should check if current active editor is in `allEnabledLanguages`. If I open VSCode on any file it will show up.

src/utils.ts
```diff
@@ -10,3 +11,33 @@ export function onWorkspaceRootChange(cb: (rootPath: string) => void): void {
...
+        window.showWarningMessage(
+            "Option 'trailingComma' as a boolean value has been deprecated. " +
+                "Use 'none', 'es5' or 'all' instead."
+        );
+    }
+    return config;
+}
+
+export function allEnabledLanguages(): DocumentSelector {
+    const config = checkConfig();
```
CiGit: Should we check it each time we change active editor ?
btw I'm pretty sure we can now remove those checks. It changed some time ago.

src/utils.ts
```diff
@@ -10,3 +11,33 @@ export function onWorkspaceRootChange(cb: (rootPath: string) => void): void {
...
+    return [
+        ...config.javascriptEnable,
+        ...config.typescriptEnable,
+        ...config.cssEnable,
+        ...config.jsonEnable,
+        ...config.graphqlEnable,
+    ];
```
CiGit: By recomputing this list each time, we will have a different behavior from the formatter. We register formatting in `activate` (only on start) but this list can change during a run (users update theirs settings). We may encounter a case were the `statusBarItem` is visible but prettier doesn't run. The opposite is also true.

CiGit:
I can merge it as is. If you don't want to change it or don't have time. Details can wait.

RobinMalfait:
I'll take a look at it in ~30minutes and I'll come back to this :)

src/utils.ts
RobinMalfait: That is very true, the reason I added it was to be a bit dynamic, if you change your settings you don't have to restart but actually you do, so maybe I don't have to compute that every single time.

src/errorHandler.ts
```diff
@@ -19,6 +21,18 @@ onWorkspaceRootChange(() => {
     outputChannelOpen = false;
 });
 
+window.onDidChangeActiveTextEditor(editor => {
```
CiGit: Shouldn't we dispose this one?

src/utils.ts
CiGit: Yup, or dynamically unregister / register our formatter :1st_place_medal:

src/errorHandler.ts
RobinMalfait: Yeah probably we need to do that

src/utils.ts
CiGit: Dynamically it can be recomputed on settings change, I don't know performance wise tho.

RobinMalfait: I have no idea how to do that to be honest. I also don't understand the vscode api docs :D

src/errorHandler.ts
RobinMalfait: Hmm, that's not that easy to do because we want to return the disposable command. So I added the `.hide()` instead of `.show()` thingy

src/errorHandler.ts
RobinMalfait: done

RobinMalfait:
You can take another look @CiGit

src/errorHandler.ts
CiGit: I think `window.activeTextEditor` and the same thing with score to decide if you show it.

src/utils.ts
CiGit: Yeah was nearly a feature when we have nothing left to do :-)

src/extension.ts
```diff
@@ -1,43 +1,13 @@
...
+import { languages, ExtensionContext, DocumentSelector } from 'vscode';
 import EditProvider from './PrettierEditProvider';
-import { setupErrorHandler } from './errorHandler';
-import { PrettierVSCodeConfig } from './types.d';
-
-function checkConfig(): PrettierVSCodeConfig {
-    const config: PrettierVSCodeConfig = workspace.getConfiguration(
-        'prettier'
-    ) as any;
-    if (config.useFlowParser) {
```
CiGit: This one is still in package.json.

src/utils.ts
```diff
@@ -1,12 +1,31 @@
...
+import { workspace, Disposable, DocumentSelector } from 'vscode';
+import { PrettierVSCodeConfig } from './types.d';
 
 let currentRootPath: string = workspace.rootPath;
 
-export function onWorkspaceRootChange(cb: (rootPath: string) => void): void {
-    workspace.onDidChangeConfiguration(() => {
+export function onWorkspaceRootChange(
+    cb: (rootPath: string) => void
+): Disposable {
```
CiGit: Good catch

src/extension.ts
RobinMalfait: It was also in other files, I removed it there as well

src/utils.ts
RobinMalfait: hehe 👍

src/PrettierEditProvider.ts
```diff
@@ -84,7 +84,7 @@ function format(
     let isNonJsParser = false;
     if (!parser) {
         // unset config
-        parser = config.useFlowParser ? 'flow' : 'babylon';
+        parser = 'babylon';
```
CiGit: I'm pretty sure we can't hit this line without having the settings warn. (enum) 
we can remove this block. 
It seems it wasn't working before. :stuck_out_tongue:

RobinMalfait: oh gotcha hold on

CiGit:
Playing around with it.
I still have one weird behavior.
Have a JS file open. Status Bar Item (SBI) shows.
Focus output or debug console SBI hides.
Focus problems or terminal SBI shows.

I really can't see how to cleanly fix this.

Note: once Eslint shows, it stays so.

CiGit:
By switching between 2 files it blinks. This is due to `toggleStatusBarItem` being called twice.
The first time with an undefined editor.

RobinMalfait:
For the jumps I do not toggle it anymore. 
When the output or debug thing is selected, we can check the "scheme"

vvs:
I am now getting compilation errors (rev 4dee32b):

```
#>npm run compile

> prettier-vscode@0.21.0 compile W:\work\github\prettier-vscode.git
> tsc -watch -p ./

src/PrettierEditProvider.ts(133,40): error TS2345: Argument of type '"babylon" | "flow" | "postcss" | "typescript" | "json" | "graphql" | undefined' is not assignable to parameter of type 'ParserOption'.
  Type 'undefined' is not assignable to type 'ParserOption'.
1:33:00 PM - Compilation complete. Watching for file changes.
```

src/types.d.ts
```diff
@@ -17,7 +17,6 @@ type TrailingCommaOption =
 export interface PrettierConfig {
     printWidth?: number;
     tabWidth?: number;
-    useFlowParser?: boolean; // deprecated
     singleQuote?: boolean;
     trailingComma?: TrailingCommaOption;
     bracketSpacing?: boolean;
```
CiGit: All those settings are in fact not optional (they will default) . removing `?` will correct compiling issue (sorry, forgot to mention that earlier) @vvs

src/errorHandler.ts
```diff
@@ -2,22 +2,54 @@ import {
...
-/**
- * Mark the outputChannelOpen as false when changing workspaces
- */
-onWorkspaceRootChange(() => {
-    outputChannelOpen = false;
-});
+function toggleStatusBarItem(editor: TextEditor): void {
+    if (editor !== undefined) {
+        if (
+            ['debug', 'output'].some(
```
CiGit: Mind adding a comment? We won't remember in some weeks !

RobinMalfait: Yeah sure that's a good idea, I was thinking about it

src/types.d.ts
RobinMalfait: Do I remove those "?" aswell in this pr or in another pr?

CiGit: Sure, it won't compile if you don't :cry: 
You didn't have compile errors?

RobinMalfait: "Debug Anyway" 😎

RobinMalfait:
@CiGit I think it is ready to be merged. Do you want me to squash everything together?

CiGit:
I've a button to do it ;-)

CiGit:
Thanks for your great work! :+1:

RobinMalfait:
Nice, thanks for merging!

CiGit:
AAAAAAAAAAAAAAAAAAAAAAAnd published. 
Thanks to reviewers too ;-)

browneye1826:
@vvs hope you are well.
Question please: You describe this perfectly [here](https://github.com/prettier/prettier-vscode/issues/161#issuecomment-317371820). Any idea how to have Prettier only show in the status bar only for enabled languages? 
For me its somehow always showing even for languages that aren't Prettier compatible (i.e. Python & Ruby).
Many thanks!

PR #191, opened 2017-08-03
Use prettier-eslint for TypeScript files

### LINKED ISSUE #190: Prettier ESLint should be invoked for TypeScript files
ESLint supports TypeScript files through [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser).

However, when `eslintIntegration` is enabled, this extension will silently skip prettier-eslint for TypeScript files: https://github.com/prettier/prettier-vscode/blob/7431dd3dc8c0e324be98378efcd1a90ce4f547fa/src/PrettierEditProvider.ts#L118

Are you happy for me to raise a PR that changes this, so TypeScript files do run through prettier-eslint?

OliverJAsh:
Fixes #190

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,9 +23,9 @@
       "dev": true
     },
     "acorn": {
-      "version": "5.0.3",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.0.3.tgz",
-      "integrity": "sha1-xGDfCEkUY/AozLguqzcwvwEIez0="
+      "version": "5.1.2",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.1.2.tgz",
+      "integrity": "sha512-o96FZLJBPY1lvTuJylGA9Bk3t/GKPPJG8H0ydQQl01crzwJgspa4AEIq/pVTXigmK0PHVQhiAtn8WMBLL9D2WA=="
     },
     "acorn-jsx": {
       "version": "3.0.1",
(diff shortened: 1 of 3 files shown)


azz:
Should probably rename `isNonJsParser` to `parserSupportsEslint` or something like that.

OliverJAsh:
@azz I can do that. Is this what is needed to get this PR merged? Any other changes?

CiGit:
Sorry, forgot that one.
What's the outcome if eslint is not configured with typescript parser?
The use case is: migrating from flow to typescript.

OliverJAsh:
@CiGit Good question.

I just tried this on my project by disabling the TS parser in my ESLint config and then invoking prettier-eslint on the file. Prettier runs fine, and the ESLint auto fixes are not applied, presumably because ESLint silently fails when trying to parse the TypeScript syntax. I think this is probably desirable?

CiGit:
This is what **I** would expect. I mostly don't want it to throw. Are configs still inferred ? indent, ...

Any other input?

Maybe rename `isNonJsParser`, it's no more what it means :-)

OliverJAsh:
@CiGit I just tested it and it does infer config still.

I have renamed and inverted `isNonJsParser` to become `doesParserSupportEslint`.

CiGit:
Could you also update prettier-eslint to 7.1 (at least), it overrides the parser. This makes more sens to me :-)
And we should be good to go

OliverJAsh:
@CiGit I can do that but that shouldn't be necessary as part of this PR, as we're already specifying TS as the parser.

OliverJAsh:
@CiGit I just tried upgrading it but the package lock file has loads of redundant changes for some reason—possibly due to an "old" npm version (I'm using 5.0.3).

CiGit:
I think I've done it with 5.4.0.
Promise, I won't look package.json too closely 😄 

I prefer if it's in this PR. The reason is it would change the output between both prettier-eslint for people having the wrong parser.

OliverJAsh:
@CiGit Done!

CiGit:
Thanks !

CiGit:
Ah, already merged, but I've just found out something. 
> prettier.eslintIntegration (default: false) - Javascript only

from the readme. contribution field in package.json should be wrong too.

OliverJAsh:
https://github.com/prettier/prettier-vscode/pull/211

Issue #194, opened 2017-08-11
Do not take over my focus when syntax error being detected


be5invis:
That popup blocks me from editing.

CiGit:
I agree with that.
It happens on the first error.
Panel opens, and is focused.
We should remove that focus.
cc @RobinMalfait

RobinMalfait:
I'll take a look at this later today/this weekend.

RobinMalfait:
Fixed in #196

PR #196, opened 2017-08-13
preserve focus when showing output channel

RobinMalfait:
fixes issue #194


CiGit:
Great. Thanks!
Short fix.

PR #211, opened 2017-09-05
Docs: update languages for ESLint integration

OliverJAsh:


CiGit:
Thanks. 

@RobinMalfait I usually press **Squash** to have 1 feature -> 1 commit. Merge to keep multiple commits.

RobinMalfait:
@CiGit Gotcha, will use that one in the future to have an even cleaner history ;)

CiGit:
It's just feeling. If a PR has multiple small feature which are not that dependent, I merge.

Side note, I won't be able to make release or code in those next weeks, I'm moving (packing unpacking physical things 😄 )

RobinMalfait:
Haha, call it coincidence! I'm moving too!

PR #215, opened 2017-09-06
Implement .prettierignore support

### LINKED ISSUE #57: Create option to ignore files (like webpack.config.js)
Hi! I'd love to have an option to include / exclude paths or files so that my webpack.config.js doesn't get ruined everytime I save the file

### LINKED ISSUE #198: Option to disable for certain folders - eg node_modules
Occaisonally you need to patch packages that have been installed (and patch-package is useful here). However, if format on save is enabled yo uget more changes that you actually want.

So it would be vey useful  to be able to stop it working in some places.

As far as I can see, VS code does not have such a feature itself.

### LINKED ISSUE #208: Doesn't respect .prettierignore

azz:
TODO:

* [x] Actually test it
* [x] Make it configurable?
* [x] ~Figure out how to properly import `ignore`~. Given up :/
* [x] Readme

Closes #208
Closes #57
Closes #198


ButuzGOL:
any info when its going to be merged and released ?

ambroselittle:
Waiting with baited breath! :D

src/ignoreFileHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+        fileWatcher,
+        fileIsIgnored(filePath: string) {
+            return ignorer.ignores(filePath);
+        },
+    };
+
+    function reloadIgnoreFile() {
+        ignorer = nullIgnorer;
+
+        if (workspace.rootPath) {
```
CiGit: This will change in future VSCode and multiple workspace.

src/ignoreFileHandler.ts
```diff
@@ -0,0 +1,46 @@
...
+            const fullIgnorePath = path.join(workspace.rootPath, ignorePath);
```
CiGit: What if ignorePath is absolute?

azz:
Thanks for the review - I'll follow up with an update in the next day/two.

src/ignoreFileHandler.ts
CiGit: For reference:
https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs

azz: Seems like it will take a fair amount of work to support settings at a `resource` level, or am I reading this wrong?

CiGit: I think all but `{LANGUAGE}Enable` settings would be quite easy. We read them on format and we have file's path (except for none persisted files).

 `ignorePath` will be a little trickier :-)
Options I see:
- With your pattern it will need to have an ignorer per workspace and find out to which workspace correspond a given file
- Read file the right file at each call (no caching). Certainly easier but also certainly worse performance wise.

mastilver:
Hi, thank you for your work on this, it will be so useful :)

My team is using both atom and vscode, so to us consistency between the two is quite important
Here's the difference between https://github.com/prettier/prettier-atom and this PR:
- doesn't look for `.eslintignore`
- `.prettierignore` path is hardcoded and relative to the root project, whereas on atom it's determined on the fly given the path of the evaluated file see: https://github.com/prettier/prettier-atom/blob/f7aaaf1270485dccc4fa5c846f5084291ebee739/dist/formatOnSave/isFilePathPrettierIgnored.js#L14-L16

If we look for the `.prettierignore` (and `.eslintignore`) from the current file, we will not need to add the `prettier.ignorePath` options

Let me know if I can help further!

azz:
Prettier itself can be configured to point to (e.g.) `.gitignore` instead of `.prettierignore`, so hardcoding the file name might be limiting. @CiGit what do you think?

mastilver:
@azz What about leave the options empty by default

Then the logic to find the ignored file would be:
1. if `ignorePath` find first `ignorePath` parents file
2. find first `. prettierignore` parents file
3. find first `. eslintignore ` parents file

mastilver:
it could be as simple as that:

```js
const path = require('path');
const findUp = require('find-up');

function getIgnoreFile(documentFileName, options) {
  const documentDir = path.dirname(documentFileName);

  return findUp.sync(options.filePath, {cwd: documentDir}) ||
             findUp.sync('. prettierignore', {cwd: documentDir}) ||
             findUp.sync('. eslintignore', {cwd: documentDir});
}

```

CiGit:
I would try to be the closest to what prettier does for this case. Not prettier-atom.

Prettier is not searching upwards for a `.prettierignore` so we shouldn't.
It's resolved in `cwd` and the closest thing we have is workspace.root I think.
You can config prettier to use an other file `ignore-path` cli only. To be consistent with our options (or prettier's API options) `ignorePath` is a good choice. It also defaults to `.prettierignore`

I don't see why we should read `.eslintignore` by default.

mastilver:
> I would try to be the closest to what prettier does for this case. Not prettier-atom.

Fair enough

> I don't see why we should read .eslintignore by default.

I just mention it as `prettier-atom` is doing it that was doing it and I was expecting the same behaviour for vscode. it also reduce ignore files.

package.json
```diff
@@ -19,7 +19,7 @@
     "url": "https://github.com/prettier/prettier-vscode/issues"
   },
   "engines": {
-    "vscode": "^1.5.0"
+    "vscode": "^1.15.0"
```
azz: I had to do this to update `vscode.d.ts` with the multi-root stuff. Is this OK?

I've maintained backwards compatibility at runtime, all I really need is the definitions.

CiGit:
`eslintignore` should be handled by setting ignorePath to it.

Even better, with eslintIntegration on. I don't know if prettier-eslint currently supports this.

azz:
Btw, code should be reviewable now but I still haven't done manual testing.

CiGit:
Thanks @azz !
I'll do it this weekend.

We should have some integration tests at some point. We start having many features.

CiGit:
@azz, I'll let you play with it first, you may see things I missed 😨 
@RobinMalfait would you review this one?

RobinMalfait:
I'll check it out. checking now!

README.md
```diff
@@ -101,6 +101,10 @@ Advanced feature. Use this to opt in / out prettier on various language ids. Res
 Use parser `graphql` for given language ids.
 Use with care.
 
+#### prettier.ignorePath (default: .prettierignore)
+Supply the path to an ignore file such as `.gitignore` or `.prettierignore`.
+Files which match will not be formatted. Set to `null` to not read ignore files. Restart required.
```
RobinMalfait: Little nitpick: `Restart required.`, of what exactly? The plugin, the editor, the window?

src/ignoreFileHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+        return getPath(workspace.rootPath);
+    }
+    if (workspace.workspaceFolders) {
+        const folder = workspace.getWorkspaceFolder(Uri.file(filePath));
+        return folder ? getPath(folder.uri.fsPath) : null;
+    }
+
+    return null;
+
+    function getPath(relativeTo: string) {
```
RobinMalfait: It works, but I am not a big fan of putting a function inside another function and below the `return` statement. Seems a bit odd to me. Maybe we can put it above the or below the `getIgnorePathForFile` function itself?

Also the only reason it works is because of hoisting if I'm not mistaken.

src/utils.ts
```diff
@@ -1,21 +1,23 @@
...
+import { workspace, Disposable, DocumentSelector, Uri } from 'vscode';
 import { PrettierVSCodeConfig } from './types.d';
 
-let currentRootPath: string = workspace.rootPath;
+let currentRootPath: string = workspace.rootPath!;
 
 export function onWorkspaceRootChange(
     cb: (rootPath: string) => void
 ): Disposable {
+    // TODO: use `workspace.onDidChangeWorkspaceFolders`
```
RobinMalfait: Is this todo still necessary or a future improvement?

src/utils.ts
```diff
@@ -1,21 +1,23 @@
-import { workspace, Disposable, DocumentSelector } from 'vscode';
+import { workspace, Disposable, DocumentSelector, Uri } from 'vscode';
 import { PrettierVSCodeConfig } from './types.d';
 
-let currentRootPath: string = workspace.rootPath;
+let currentRootPath: string = workspace.rootPath!;
```
RobinMalfait: General question, what is the `!;` at the end, and what does it do and why do we need it?

> It's probably my lack of TypeScript knowledge, but can't easily find what it is.

README.md
CiGit: All of them in fact.

We have same annotation to other settings ({LANG}Enabled)

src/ignoreFileHandler.ts
CiGit: Yup hoisting.
Could be a used once helper function.

src/utils.ts
CiGit: It means 'I, as a coder, assert: it **IS** defined'

src/utils.ts
CiGit: I think it is for VSCode's multiple workspaces support.

README.md
RobinMalfait: Allright, than we can keep this.

src/utils.ts
RobinMalfait: oh, til!

src/ignoreFileHandler.ts
RobinMalfait: Are you okay with this or do you want to see it moved? Otherwise I'm going to dismiss/accept my request.

CiGit: Na it could definitely be moved at least before. It's of course nicer :-)

I still will let @azz test his own work before merging.

src/utils.ts
azz: Yeah I don't want to tackle too much in this PR.

azz:
Sorry I took so long to get to this, somehow I missed the review notifications.

I've removed the hoisting mentioned above and ran through a bunch of manual test cases and everything looks fine to me.

Cases tested:
* No `.prettierignore` file, everything is formatted.
* An empty `.prettierignore` file, everything is formatted.
* Change `.prettierignore`, files matching patterns are not formatted.
* Rename to `.prettier_ignore`, everything is formatted.
* Change setting `"prettier.ignorePath" to `.prettier_ignore`, restart, correct file is detected.

RobinMalfait:
You have my go, @CiGit 👍

CiGit:
Thanks @azz !!

If no one is against it (including @azz), I'd like to add @azz as a `Collaborator` to this project 😃 

Lot of work done in some PR / issues.

CiGit:
I've run the same tests, but with an absolute path at your last point.

side note, I've started setting up some tests on my machine. PR coming soon :-)

Issue #207, opened 2017-08-31
Provide integration wit stylelint


vvs:
Now that prettier supports CSS/SCSS formatting, it would be really nice to integrate prettier-vscode with stylelint in a manner similar to what is done with eslint integration for JavaScript files.

Ideally, there should be a config option to enable/disable this integration (just like with eslint), so that the SCSS/CSS styles formatted with prettier are then auto-fixed via stylelint to confirm to stylelint rules.

Stylelint is adding more and more support for its `--fix` functionality.

Here's one of the use cases I have: The source base I am working on has a stylelint rule that all color HEX values must be uppercase. Prettier formats the HEX values to be lowercase, which breaks stylelint checks.

And since prettier is an 'opinionated' formatter, it doesn't provide a way to configure this behavior.

CiGit:
I'm not sure this would bring something more than only relying on stylelint in this case. Disclaimer: I don't know stylelint.

Is there any project we could start of? (prettier-stylelint, ...)

vvs:
@CiGit Stylelint is very simialr to eslint, it checks for common problems in styles, and it also enforces some of the stylistic rules on how to format the CSS/SCSS files.

For the list of stylelint rules, take a look here: https://stylelint.io/user-guide/rules/

The rules that can be fixed by stylelint (`--fix` option) are marked as Autofixable. As you see, Stylelint doesn't really reformat the whole files, it mostly does gentle things, like cases, empty lines, indents (some of these options are not available in prettier, which is probably not that bad in common case, but pretty bad in those cases when you are working on big projects and cannot change the rules and force everybody switch to prettier).

Without this prettier/stylelint integration I can't really use prettier (and this plugin in VS Code) in some projects, since the SCSS output prettier produces basically breaks the build due to stylelint errors. With javascript files this is not a problem, since I can rely on prettier-eslint integration to fix the differences.

I am not aware of any current project like `prettier-stylelint` at the moment.

hugomrdias:
Working on this, already have a basic version of a  prettier-stylelint tool a few for tests and I'll do a PR here adding support for it.

CiGit:
This seems to be done since a long time ...

PR #218, opened 2017-09-12
feat: initial support for prettier-stylelint

hugomrdias:
this adds initial support for [prettier-stylelint](https://github.com/hugomrdias/prettier-stylelint), im still adding features but the basic stuff at least for the extension works well.

please keep in mind that i don't do ts or vscode extensions much :) be gentle 

ref #207

### DIFF
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -110,10 +110,24 @@ export function addToOutput(message: string): void {
  * @returns {string} formatted text or defaultText
  */
 export function safeExecution(
-    cb: () => string,
+    cb: (() => string) | Promise<string>,
     defaultText: string,
     fileName: string
-): string {
+): string | Promise<string> {
+    if (cb instanceof Promise) {
+        return cb
+            .then(returnValue => {
+                updateStatusBar('Prettier: $(check)');
+                return returnValue;
+            })
+            .catch((err: Error) => {
+                addToOutput(addFilePath(err.message, fileName));
+                updateStatusBar('Prettier: $(x)');
+
+                console.log(err);
+                return defaultText;
+            });
+    }
     try {
         const returnValue = cb();
 
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -60,7 +60,7 @@ export function registerDisposables(): Disposable[] {
  * 
  * @param message The message to put inside the statusBarItem
  */
-function updateStatusBar(message: string): void {
+export function updateStatusBar(message: string): void {
     statusBarItem.text = message;
     statusBarItem.show();
 }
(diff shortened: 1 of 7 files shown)


CiGit:
Thanks for that PR and your prettier-stylelint.
Could your prettier-stylint format be synchronous ? To be able to use `safeExecution` without having to export everything ?

hugomrdias:
stylelint doesnt have a sync api, i can create a safeExecutionAsync instead and leave everything abstracted in the utils

CiGit:
This is also a solution. Make `safeExecution` handle both.

hugomrdias:
@CiGit does this seem good to you ?

CiGit:
Yup, I prefer. asking @RobinMalfait  for a review of `safeExecution`

About prettier-stylelint, which prettier/ stylelint version does it pickup?

src/errorHandler.ts
```diff
@@ -110,10 +110,24 @@ export function addToOutput(message: string): void {
...
+        return cb
+            .then(returnValue => {
+                updateStatusBar('Prettier: $(check)');
+                return returnValue;
+            })
+            .catch((err: Error) => {
+                addToOutput(addFilePath(err.message, fileName));
+                updateStatusBar('Prettier: $(x)');
+
+                console.log(err);
```
RobinMalfait: Maybe we need to remove this line

RobinMalfait: Hmm, now that I'm thinking about it, the `safeExecution` has a lot of duplicated code.

hugomrdias: Ofc damn .. I'll remove that

hugomrdias:
@cigit for now the bundled version, but I'll be fixing that in a near future

hugomrdias:
@RobinMalfait pushed a commit removing the console log

CiGit:
My concerns are how often do you expect it to be updated, until it has all the feature you'd like ?
On your repo I see 9 releases in 8 days. I don't think we will be able to follow so many releases here :-)

hugomrdias:
i dont expect to have that many releases from now on. besides ill keep sending PR's here for each new release.
plus after i add support for relative package lookup (which im already working on) this extension need to have the latest version bundled to work properly

hugomrdias:
@CiGit with this last commit your concerns should be addressed

CiGit:
Please resolve conflicts, I'll merge after.

CiGit:
And update the readme please :-)

CiGit:
Thanks !

WickyNilliams:
Is there some way to configure this to run over JS files? Use case: styled-components.

The regular stylelint extension for vscode has a config option `additionalDocumentSelectors` so you can configure it to run over JS files as well as the usuall css/scss/less. eg:

```js
  "stylelint.additionalDocumentSelectors": [
    "javascriptreact",
    "javascript"
  ]
```

Happy to make an issue in the appropriate repo, but thought i'd ask here given this was the integration point

PR #231, opened 2017-10-02
Initial integration tests

CiGit:
Currently this only tests if it formats correctly (ie same output as prettier). (prettier only)

I could not find a way to test graphql.

 - [x] automatic (travis?)
 - [x] Doc


.prettierignore
```diff
@@ -1,2 +1,3 @@
 package.json
-package-lock.json
\ No newline at end of file
+package-lock.json
+testProject/
```
CiGit: The purpose of the test is to format ugly code.

testProject/.vscode/settings.json
```diff
@@ -0,0 +1,31 @@
+{
```
CiGit: Restoring default values

test/index.ts
```diff
@@ -0,0 +1,23 @@
...
+// a possible error to the callback or null if none.
+
+var testRunner = require('vscode/lib/testrunner');
+
+// You can directly control Mocha options by uncommenting the following lines
+// See https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically#set-options for more info
+testRunner.configure({
+    ui: 'tdd', // the TDD UI is being used in extension.test.ts (suite, test, etc.)
+    useColors: true, // colored output from test results
+    timeout: 10000,
```
CiGit: It may take some time depending on the machine, Increased timeout from 2s (default) to 10s (randomly selected value)

src/types.d.ts
```diff
@@ -70,7 +70,7 @@ interface ExtensionConfig {
  */
 export type PrettierVSCodeConfig = ExtensionConfig & PrettierConfig;
 export interface Prettier {
-    format: (text: string, options?: PrettierConfig) => string;
+    format: (text: string, options?: Partial<PrettierConfig>) => string;
```
CiGit: No option is required

CiGit:
MacOS travis may be slow to start.
Does it make sens to also test on it?

CiGit:
Ready to review, I won't do more in that PR

package.json
```diff
@@ -228,14 +228,14 @@
     "vscode:prepublish": "tsc -p ./",
     "compile": "tsc -watch -p ./",
     "postinstall": "node ./node_modules/vscode/bin/install",
-    "test": "node ./node_modules/vscode/bin/test",
+    "test": "CODE_TESTS_WORKSPACE=testProject node ./node_modules/vscode/bin/test",
```
azz: Suggest using [`cross-env`](npmjs.com/package/cross-env) here.

CiGit: Right, correction done.

CiGit:
Seems we missed a Changelog somewhere ...
> Warning: `parser` with value "postcss" is deprecated. Use "css", "less" or "scss" instead.

PR #240, opened 2017-10-13
Add a note about custom keybindings to readme

kostasmanionis:
I was looking for a way to rebind `CMD + Shift + P` and it wasn't immediately obvious that I can rebind  `editor.action.formatDocument` and `editor.action.formatSelection`. So decided to add a little note to the readme 🤔


CiGit:
Thanks !

Issue #232, opened 2017-10-02
Custom Prettier Resolution


bradenhs:
Instead of needing to have prettier installed in your current folder's `node_modules` directory it'd be nice to have the option to specify where you'd like the extension to look for your prettier installation. Resolution strategy would be as follows:

1. If "prettier.path" is defined try to use that.
2. See if there's a copy in the current directory's node_modules folder.
3. Use a copy of prettier bundled with the installation.

The issue I ran into: I'm using yarn workspaces and it hoists dependencies if they're shared between different workspaces. Because of this my prettier installation isn't in the individual workspace's node_modules folder. It'd be nice to have the ability to set the following in my `.vscode/settings.json` file:

```
"prettier.path": "../../node_modules/prettier"
```

Thoughts?

azz:
Seems totally reasonable to me. I think typescript has a similar setting.

CiGit:
If I understood correctly your file system looks like

- node_modules/prettier
- sub/project/package.json{"prettier"}

I think we can update our local resolution algorithm to resolve this case. Resolve from package.json's folder.

:heavy_plus_sign: No new settings (and we wouldn't have to do that for other modules we have)

:heavy_minus_sign:  Still won't resolve a randomly installed prettier (global install, ...)

bradenhs:
I'd be just as happy with either solution.

priley86:
+1

Bnaya:
Same issue here, with mono repo
```
Failed to load prettier from /Users/.../project-root/packages/webapp/node_modules/prettier. Using bundled
```
yarn workspace is hoisting `prettier` to `project-root/node_modules`

agoldis:
another use case that needs that - prettier is hoisted as part of lerna configuration, what left for lerna sub-package is only symlink to hoisted npm module

kevin-ashton:
Any update on this?  I think the original proposal seems ideal whereas it provides max flexibility. 

As I understand it I don't think the proposal of  updating the local resolution algorithm would work on my project. We have a cli tool (that can live anywhere) that handles running various things across projects for us including formatting code. It would be very helpful to be able to specify the path of the prettier executable similar to what the SublimeText plugin can do.

kevin-ashton:
I think creating an algorithm to try and intelligently determine which prettier module to use is reasonable. That being said if you don't have a way of overriding the prettier resolution you are restricting this tool projects that happen to follow the organization patterns your algorithm is expecting.

[...]

PR #241, opened 2017-10-13
Fix infinite recursion with top-level package.json

### LINKED ISSUE #138: Maximum call stack size exceeded on JavaScript file format
- VSCode Version: 1.13.1 (379d2efb5539b09112c793d3d9a413017d736f89)

- OS : Windows 10 Pro

Steps to Reproduce:

1. Open JavaScript file with wrong indentation
2. Hit Shift+Alt+F
3. Maximum call stack size exceeded

I tried disabling each and every extension in VSCode and then enabling them one by one and the only one that gave me that error is this extension.

azz:
Fixes #138

### DIFF
--- a/src/requirePkg.ts
+++ b/src/requirePkg.ts
@@ -1,6 +1,6 @@
 import { addToOutput } from './errorHandler';
 
-const path = require('path');
+import * as path from 'path';
 const readPkgUp = require('read-pkg-up');
 
 /**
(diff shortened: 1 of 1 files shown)


azz:
I couldn't seem to find a `sindresorhus` that does this. Maybe using [`resolve`](https://www.npmjs.com/package/resolve) instead would be better?

CiGit:
I first implemented something like `resolve`, querying for a module.

Troubles we had, if a package installed by a user had a dependency on `prettier` we would load it (due to module flattening) . That's why this second version was checking for an explicitly installed prettier.
Seems `resolve` gives a `pkg` info in the async version.

It may also correct #232

PR #246, opened 2017-10-20
Updated description of 'prettier.trailingComma'

prog-rajkamal:
## Problem:

While editing prettier settings in vscode settings page, the description of property 'prettier.trailingComma'  is lacking. 

The original description was not intuitive and lacked the list of available option. (this is somewhat mitigated by vscode intellisense)

## Resolution:

Updated the description to be same as in README

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
             "all"
           ],
           "default": "none",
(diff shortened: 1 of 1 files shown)


CiGit:
Thanks you!

Issue #168, opened 2017-07-16
no more popup warnings :(


smooJitter:
I use to get this nice popup warnings when I would try to save a file with critical errors. I upgrade prettier-vscode and I did change a few settings (I'm not sure what, but I no longer get the nice popup warnings.  When there's error the file would save but prettier will not reformat the code.  Pretter only reformats the code when there are areas.  Although this too is a nice feature, I miss the popups directing me to the error.  How can I get the popup warning back?

RobinMalfait:
The popups were annoying for many users. There is a new output channel where the errors are posted. There is also a statusbar item which indicates if there are errors or not.

Output channel:

![](http://d.rbn.nu/NYFasg+)

Statusbar item:

![](http://d.rbn.nu/zQRSM+)

smooJitter:
lol Understood.  I'll miss them.

alexozer:
@RobinMalfait Would it be reasonable to add an option to disable the Output from prettier? It seems jarring to have eslint problems reported subtly with squiggly lines in the editor but have prettier use Output.

orta:
I shipped a PR to disable the output

Issue #243, opened 2017-10-13
Multi-root Workspaces and Prettier


seanmcbreen:
Hi,

I'm from the VS Code team and I'm reaching out with some additional resources for adopting multi-root workspaces.

But first let me say thank you for working on and maintaining [Prettier This for VS Code](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) which is one of the most popular extensions, we really appreciate the work you have done to make VS Code better.

The purpose of this issue is that we are very close to releasing a significant new feature for VS Code - Multi-root Workspaces - which is the [most requested](https://github.com/Microsoft/vscode/issues?q=is%3Aopen+is%3Aissue+label%3Afeature-request+sort%3Areactions-%2B1-desc) missing feature. If you have not heard of multi-root workspaces we have [described this in several of our recent updates](https://code.visualstudio.com/updates/v1_17#_preview-multi-root-workspaces).  Today it's only enabled in our insiders release but **our intent is to enable it for all users** in our [next update](https://github.com/Microsoft/vscode/issues/35875), targeted for release in the first week of November.

This is a pretty significant change and over the last four months we have updated pretty much all of VS Code to be Multi-root aware as well as updated all of our own extensions.  In the process of doing this we have created a set of [guidelines for Extension Authors](https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs) to follow to ensure their extension can fully leverage the awesomeness that is Multi-root workspaces.

**Our Request:**

Having looked at your extension we think changes are required in how you:

* Make use of the (now deprecated) `workspace.rootPath` property, there is a proposed migration path and sample for this case.
* Contribute or change settings, specifically you need to review whether some of the settings can be applied on a resource (= file location) level instead of being global (= workspace level).

**How Can We Help/Resources:**

We realize we are asking for your help here - our goal is to make sure that your extensions works well with multi-root and continues to be loved by the community.  [The guide above outlines the changes that are required](https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs) and we have a [samples](https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs#samples) as well.

[...]

PR #247, opened 2017-10-22
Add an option to suppress popping up the output panel for a failed run

orta:
Should be enough to close https://github.com/prettier/prettier-vscode/issues/168

Allows a user to opt out of the output channel being popped up, TBH it might be worth thinking about making this be the default and allow opting in to showing the output channel on an error. I can't imagine that the errors are inside prettier any more, and I'm willing to bet that VS Code makes it pretty obvious that your source code isn't compiling correctly. Plus you'll see the "prettier x" in the bottom right.


CiGit:
Hey, thanks for your PR.

I don't think an option for that is a good move.
We could discuss to remove the automatic opening. It's definitely not in prettier's scope to lint your code. `prettier : X` may be enough to say something went wrong
I'll let @RobinMalfait decide on that one ;-)

As I'm working on #243 there are things which will have to change anyway. (Or I'll have to do some dirty checks to keep the same functionalities)

RobinMalfait:
Hey! Thanks for the PR!

I have been annoyed by the popup itself, especially because it doesn't close when the formatting _is_ correct.

The changes I have requested should result in remove some code (Yay!)

Thanks for the PR!

RobinMalfait:
Perfect!
@CiGit can you verify? Code LGTM

CiGit:
@RobinMalfait 
>Always write to the output log (I think this is already the case)

did you mean https://github.com/prettier/prettier-vscode/blob/master/src/PrettierEditProvider.ts#L161 ?
`onWorkspaceRootChange` is quite embarrassing for #243

RobinMalfait:
I was talking about https://github.com/prettier/prettier-vscode/blob/844ecd6820e6ef95bf432dac761fcd0fd4757520/src/PrettierEditProvider.ts#L158

I didn't know the warning message was still used?

CiGit:
Well it seems to be used. And It can be removed I think.

orta:
^ I can't tell if this is you two asking me to amend this PR

RobinMalfait:
@orta Go ahead 😄

RobinMalfait:
@CiGit Cool, thanks for the PR @orta !

PR #252, opened 2017-11-03
Multi root (vscode next)

CiGit:
Adapt settings for resource usage. ie settings which are relevant to a file.

remove rootpath usage.

Will need to make test on an insider version. (Welcome testers !)

fix #243 

Edit: Made some tests on insider with 2 folders. Was working 😄

Issue #229, opened 2017-09-27
feature request: only format if prettier config is found (as in prettier-atom)


jokeyrhyme:
OMG, prettier plugins are so amazing!

The Atom plugin for Prettier has this neat option:
"Only format if a Prettier config is found"
https://github.com/prettier/prettier-atom/blob/v0.40.0/src/config-schema.json#L131

It would be terrific if this plugin for VSCode had a similar feature

CiGit:
There is a discussion on that in #199 
We should merge this together.

jokeyrhyme:
Thanks for the link! (I did check but my search terms were different)

Issue #199, opened 2017-08-21
Enable only if prettier is installed


jantimon:
Would it be possible to add an option which disables this extension if prettier is not installed in the root package json?

CiGit:
There are already some ways to disable the extension. I'm not sure one more would really bring something.
- Disable the extension for the workspace.
- Empty all arrays from `prettier.{LANGUAGE}Enable` settings.

nilshartmann:
I'd also really like to have this option: As I'm using prettier only in very few of my workspaces, some "opt-in behaviour" would be much smoother for my work.

RobinMalfait:
if you are only using it in very few projects/workspaces. Then you can disable it by default, and enable it where necessary.

CiGit:
In my setup I can't enable for a given workspace. It is grayed out.

But a similar behavior can be accomplished:
User settings : `prettier.javascriptEnable:[]`

workspace settings where needed: `prettier.javascriptEnable:["javascript", "javascriptreact"]`

jantimon:
Well - I have like 20 plugins and many open source and internal projects I work on.
If I know would have to add/remove/configure all 20 for all of my projects I would rather lose productivity. Which sounds funny for a plugin which should add more productivity.

Many awesome linting tools and similar plugins already do that and I guess the technical solution is quite easy to do.

RobinMalfait:
I'm confused now. You want an option to configure the extension so that it is enabled/disabled. But you don't want to configure the extension because you have too many projects?

jantimon:
Ah okay sorry for all the confusion :)

My idea is as follows:

1. install prettier-vscode
2. activate prettier-vscode for all projects
3. add only one global config setting like `prettier.smartEnable = true` which would run prettier only if the active project has a package.json including prettier

This would reduce the amount of project specific configuration for many plugin users.

RobinMalfait:
I know that @CiGit is not a big fan of options. I don't care that much about all the options.
What are your thoughts @CiGit?

Maybe the option can be long and descriptive: `prettier.disableWhenThereIsNoLocalPrettierVersionInstalled` (Naming is hard, isn't it)

jthegedus:
I would rather an option like this:
`prettier.disableUnlessPrettierConfigIsPresent`

CiGit:
I really want to avoid options. And even more when they seem to add something which is already possible. 
Maintaining options is hard.
Conflicting options is a nightmare.

[...]

PR #67, opened 2017-04-08
Require prettier to be explicitly installed

CiGit:
Changing prettier's resolution algorithm.
Instead of searching for a `node_modules/prettier` search recursively for a package.json with prettier as dependency or devDependency and then try to load it from `node_modules`.

Pulls in a new package [read-pkg-up](https://www.npmjs.com/package/read-pkg-up)

fixes #65 
fixes #59

PR #256, opened 2017-11-07
Use prettier@1.8.0 to remove *Enable settings

### LINKED ISSUE #237: .prettierc overrides are ignored
It seems the plugin ignores prettier's overrides options. E.g. the following `.prettierc` fails for the directory `build_tools`:

```yaml
{
  "trailingComma": "all",
  "overrides": [
    {
      "files": "build_tools/*.js",
      "options": {
        "trailingComma": "es5"
      }
    }
  ]
}
```

It works fine if I run prettier from the command line.

azz:
This is obviously a breaking change but it greatly simplifies the process of supporting new languages (only bumping bundled prettier is required now).

Can't really see a reason to keep the `*Enable` settings any more...

We could do a 1.0.0 release with this and #252.

Fixes #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+- Removed all `*Enable` settings, these are now inferred from Prettier itself. Use scoped `editor.formatOnSave` to disable formatting some languages. (See README)
+- Markdown support
+- Prettier 1.8
 
 ## [0.24.0]
 - new setting, ignorePath. Ignore files.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
-- Removed all `*Enable` settings, these are now inferred from Prettier itself. Use scoped `editor.formatOnSave` to disable formatting some languages. (See README)
+- Removed all `*Enable` settings, these are now inferred from Prettier itself. Use scoped `editor.formatOnSave` to disable formatting some languages on save.
+(See README)
 - Markdown support
-- Prettier 1.8
+- Prettier 1.8.2
 
 ## [0.24.0]
 - new setting, ignorePath. Ignore files.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prettier-vscode",
-  "displayName": "Prettier - JavaScript formatter",
+  "displayName": "Prettier - Code formatter",
   "description": "VS Code plugin for prettier/prettier",
   "version": "0.24.0",
   "publisher": "esbenp",
@@ -32,7 +32,7 @@
   "contributes": {
     "configuration": {
       "type": "object",
-      "title": "Prettier - JavaScript formatter configuration",
+      "title": "Prettier - Code formatter configuration",
       "properties": {
         "prettier.eslintIntegration": {
           "type": "boolean",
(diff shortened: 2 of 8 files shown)


CiGit:
Initially I've seen `*Enable` as a way to disable prettier for a given language, to be able to use any other formatter. Or to add some *custom* languages which are supported by prettier (renamed language ids like `JSX` <-> `javascriptreact`)
It also allows some tricks like: https://github.com/prettier/prettier-vscode/issues/199#issuecomment-342382635

I'm really pleased if we can remove those `*Enable*`
I think we should do #199 before releasing next version with those changes.
v1 seems fine!

Sorry no time to review right now.

CiGit:
travis seems to have some DNS issues with their mac box.

> Could not resolve host: github.com

> The maintenance work performed earlier today seems to have left some unresolved networking problems. No OSX builds are currently passing. We are investigating the situation.
Posted about 1 hour ago. Nov 09, 2017 - 13:06 UTC

src/utils.ts
```diff
@@ -21,13 +21,22 @@ export function getConfig(uri?: Uri): PrettierVSCodeConfig {
 }
 
 export function allEnabledLanguages(): DocumentSelector {
-    const config = getConfig();
+    return (require('prettier') as Prettier)
+        .getSupportInfo()
+        .languages.reduce(
+            (ids, language) => [...ids, ...language.vscodeLanguageIds],
+            []
+        );
```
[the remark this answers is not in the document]
azz: Not quite, this is `flatMap`.
```js
> x = [{a:[1]}, {a:[2]}, {a:3}]
> x.reduce((ids, obj) => [...ids, ...obj.a], [])
[ 1, 2, 3 ]
> x.map(obj => obj.a)
[ [ 1 ], [ 2 ], [ 3 ] ]
```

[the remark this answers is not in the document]
CiGit: My bad, missed the spread part 😕

src/extension.ts
```diff
@@ -1,21 +1,17 @@
...
     const editProvider = new EditProvider(fileIsIgnored);
     const languageSelector = allEnabledLanguages();
 
-    // CSS/json/graphql doesn't work with range yet.
-    const rangeLanguageSelector: DocumentSelector = [
-        ...config.javascriptEnable,
-        ...config.typescriptEnable,
-    ];
+    // Range selection is only supported for JS/TS
+    const rangeLanguageSelector = allJSLanguages();
```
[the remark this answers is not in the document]
azz: We should add range selection capability to `getSupportInfo`.

package.json
```diff
@@ -243,9 +139,8 @@
   "dependencies": {
     "prettier-stylelint": "^0.4.1",
     "ignore": "^3.3.5",
-    "prettier": "1.7.3",
+    "prettier": "1.8.2",
```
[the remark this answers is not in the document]
azz: Looks like `requirePragma` is missing, too.

What about removing these options too, in favour of `.prettierrc`?

azz:
> Is there any reason you prefer indexOf over includes ?

Nope, was just using it because I forgot this extension doesn't have to worry about node v4 compat.

azz:
> Use case: I'd like to use prettier for JS but not for CSS. For CSS I'd like to use placeholder

How about a blacklist instead then?

```json
"prettier.disableLanguages": ["css"]
```

package.json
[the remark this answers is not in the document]
CiGit: > What about removing these options too, in favour of .prettierrc?

Which options? All of them ?
I see them as useful when you have no project.
`requirePragma` seems to be useful only when in a project tho.

CiGit:
> Can .prettierignore be used instead?

This would be really neat !!
Sadly I think this won't be possible with multi-root. We register one formatter but may have multiples prettierignore (1 per folder) 

The blacklist is the only (?) thing left I think

Maybe we should delay such implementation. See if someone asks for it.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file.
 
 ## [Unreleased]
+- Removed all `*Enable` settings, these are now inferred from Prettier itself. Use scoped `editor.formatOnSave` to disable formatting some languages. (See README)
```
CiGit: I don't see formatOnSave as a synonym for disable.

src/PrettierEditProvider.ts
```diff
@@ -79,27 +57,27 @@ async function format(
...
+    const dynamicParsers = getParsersFromLanguageId(languageId, localPrettier.version);
+    let useBundled = false;
+    let parser: ParserOption;
+
+    if (!dynamicParsers.length) {
+        const bundledParsers = getParsersFromLanguageId(
+            languageId,
+            bundledPrettier.version
+        );
+        parser = bundledParsers[0] || 'babylon';
```
CiGit: In which case would it be the fallback?

CiGit:
If you are happy with your PR, you can merge it 😉

src/PrettierEditProvider.ts
azz: Hopefully never!

azz:
> Sadly I think this won't be possible with multi-root. We register one formatter but may have multiples prettierignore (1 per folder) and using prettierignore on format doesn't allow the user to register any other formatter. Well technically they can install an other formatter and only one of those would run. (never exactly understood which one)

https://code.visualstudio.com/docs/extensionAPI/vscode-api#languages.match
> Multiple providers can be registered for a language. In that case providers are sorted by their score and the best-matching provider is used. Failure of the selected provider will cause a failure of the whole operation.

Which sucks, because I think that means we can't support multi-root language disabling at all.

package.json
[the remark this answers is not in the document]
lipis: Yes, let's not remove all of the options.. as already mentioned they are useful when there is no project :)

lipis:
Are we there yet?!

(Don't mean to rush anything.. but it would be nice to start formatting markdown inside the editor 😄)

azz:
Been super busy lately. Hopefully get back to this in the next few days.

azz:
I think this is pretty much ready to go - we can look at ignore support in a focussed PR rather than adding it to this one.

src/PrettierEditProvider.ts
```diff
@@ -48,6 +49,10 @@ async function format(
     ) as any;
     const localPrettier = requireLocalPkg(fileName, 'prettier') as Prettier;
 
+    if (vscodeConfig.disableLanguages.includes(languageId)) {
+        return text;
+    }
```
CiGit: Instead of a noop we should just not register the formatter for those languages. To be able to use an other formatter for instance.

in extension.ts:
`languageSelector - disableLanguages`
`rangeLanguageSelector - disableLanguages`

This means "restart required" :smile:

src/PrettierEditProvider.ts
```diff
@@ -92,7 +97,8 @@ async function format(
             parser: parser,
             semi: vscodeConfig.semi,
             useTabs: vscodeConfig.useTabs,
-        },
+            proseWrap: vscodeConfig.proseWrap,
+        } as PrettierConfig,
```
CiGit: Was there a problem without the cast?

src/PrettierEditProvider.ts
CiGit: Ok, never mind, forgot multi-root......

lipis:
Approved?!!

lipis:
and now we wait.. and now on refresh mode: https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

CiGit:
Will publish it a little later, I'd like to merge #252 first.

src/PrettierEditProvider.ts
CiGit: In fact, It could be a `resource: "window"` scope settings. Like our old `*Enabled` settings.

azz:
Shouldn't this have been published as 1.0.0 rather than 0.26.0?

src/PrettierEditProvider.ts
azz: There was no type-checking on it before. Now it will be an error if incorrect properties are added

src/PrettierEditProvider.ts
azz: IMO `"resource": "window"` should not exist because it doesn't do anything in multi-root. Seems annoying that there's no way to skipping formatting on a per-resource level.

Maybe a feature request to `return null` from a formatter to skip it in VSCode?

lipis:
Lost opportunity for the tweet :) https://twitter.com/PrettierCode/status/933269965830082560 people like new versions especially when they reach 1.0.0

src/PrettierEditProvider.ts
azz: https://github.com/Microsoft/vscode/issues/38945

CiGit:
I'd like to put `requireConfig` with 1.0.0 #67 #229.
Some long standing feature requests.

src/PrettierEditProvider.ts
CiGit: On VSCode's side might be a good thing. :+1:

azz: It actually seems to work if you throw an error... https://github.com/Microsoft/vscode/issues/38945#issuecomment-346319330

```ts
    if (vscodeConfig.disableLanguages.includes(languageId)) {
        throw new Error('Skip me!');
    }
```

CiGit: Seems great!  just waiting for a confirmation on that *functionality*. 
API says:
>Returns: A set of text edits or a thenable that resolves to such. The lack of a result can be signaled by returning undefined, null, or an empty array.

PR #262, opened 2017-11-13
Fix settings file typo in README

kostasmanionis:
Looking at the [docs](https://prettier.io/docs/en/configuration.html), it shouldn't have that dot

https://github.com/prettier/prettier-vscode/blob/221e65c8e245aaea02dc4deb2a566680c3f73caa/src/configCacheHandler.ts#L11

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Settings will be read from:
 1. File system, first matching file in
     1. `package.json` - `prettier` key
     1. `.prettierrc`
-    1. `.prettier.config.js`
+    1. `prettier.config.js`
 1. VSCode prettier's settings, described below
 1. VSCode prettier's default settings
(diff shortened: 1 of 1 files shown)


twhid:
FWIW, I couldn't get the config to load in vscode Version 1.17.2 (1.17.2) unless it was yaml. Could be a prettier bug, kept getting a parse error:
```
Unexpected token, expected ; (/path/to/project/.prettierrc:2:21)
  1 | {
> 2 |     "bracketSpacing": true,
    |                     ^
  3 |     "jsxBracketSameLine": false,
  4 |     "parser": "babylon",
  5 |     "printWidth": 100,
```

RobinMalfait:
Thanks for the contribution!

PR #268, opened 2017-11-21
remove osx from travis

CiGit:
I'd like to remove OSX tests from running on travis.

They are slow!
And I don't think they bring a lot to the table. We make no system call...


CiGit:
Doesn't seem to be controversial. Merging

PR #283, opened 2017-12-05
prettier 1.9 and config handling

### LINKED ISSUE #81: Respect editorconfig specs
As stated in https://github.com/prettier/prettier/issues/42 and https://github.com/prettier/prettier-atom/issues/90, it would be great to have prettier surrender to the specs defined in a project-wise `.editorconfig` file.

### LINKED ISSUE #199: Enable only if prettier is installed

### LINKED ISSUE #248: Do not use editor configuration if project uses prettier
Prettier VSCode shouldn't merge the custom editor configuration with the configuration from the projects as this leads to a different style than the one from the CLI. It would be great if the custom configuration would be only used when the project doesn't use Prettier directly.

### LINKED ISSUE #273: enable stylelintIntegration with scss not work with 'prettier-stylelint'?
i am confused for  a long time, because i can't make my editor to format my scss file crollectly even by this extension.

i try myself change my vscode config for the extension so many time but not work;

i want to format my stylesheet file by my stylelint config.

but i found that when i try to launch the extension and debug that:
https://github.com/prettier/prettier-vscode/blob/8f31fa2c2af98087d9df50f04b15100f0895a303/src/PrettierEditProvider.ts#L111

why condition include `parser === 'postcss'`,   `parser` was equaled 'scss' when i launched and debugged with scss file?   

how can i change the prettier config to make it work?

CiGit:
* proseWrap config change
* Don't merge vscode's option in prettierConfig -- fix #248 
* dependencies bump
* Fix #273 add style parsers
+ arrowParens config
+ requireConfig config. Format only files with a prettierconfig -- fix #199
+ using resolveConfig + editorconfig -- fix #81 

- [x] Readme
- [x] Changelog
- [ ] ~~json validation for `prettierrc.json`~~ other PR

## Break

Don't handle deprecated options. Workaround: prettierconfig

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,12 @@
           "description": "Use 'prettier-stylelint' instead of 'prettier'. Other settings will only be fallbacks in case they could not be inferred from stylelint rules.",
           "scope": "resource"
         },
+        "prettier.requireConfig":{
+          "type": "boolean",
+          "default": false,
+          "description": "Require a 'prettierconfig' to format",
+          "scope": "resource"
+        },
         "prettier.ignorePath": {
           "type": "string",
           "default": ".prettierignore",
@@ -128,9 +134,20 @@
           "scope": "resource"
         },
         "prettier.proseWrap": {
-          "type": "boolean",
-          "default": true,
+          "type": "string",
+          "enum": ["preserve", "always", "never"],
+          "default": "preserve",
           "description": "(Markdown) wrap prose over multiple lines"
+        },
+        "prettier.arrowParens": {
+          "type": "string",
+          "enum": [
+            "avoid",
+            "always"
+          ],
+          "default": "avoid",
+          "description": "Include parentheses around a sole arrow function parameter",
+          "scope": "resource"
         }
       }
     },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -157,18 +174,18 @@
...
+    "cross-env": "^5.1.1",
     "mocha": "^3.5.3",
-    "typescript": "^2.4.2",
-    "vscode": "^1.1.6"
+    "typescript": "^2.6.2",
+    "vscode": "^1.1.10"
   },
   "dependencies": {
+    "ignore": "^3.3.7",
+    "prettier": "1.9.0",
```
azz: 1.9.1 is out now

RobinMalfait:
Could you also bump the version of prettier-eslint if there is a new version?

package.json
```diff
@@ -157,18 +174,18 @@
...
+    "prettier-eslint": "^8.2.4",
```
CiGit: @RobinMalfait  8.2.0 -> 8.2.4
There is also a 8.2.5 now

RobinMalfait: Whoops, didn't see that!

CiGit:
Deprecated old options, notably "postcss" and updated stylelint check with new styles

CiGit:
Good for a review !

CiGit:
Does someone see something more we would need for a 1.0.0 release?

Issue #292, opened 2017-12-12
Is prettier.eslintIntegration a setting that can be set in the .prettierrc file?


duro:
I tried setting it in .prettierrc (rather than in VSCode settings) and it did not seem to work. Does work when set in VSCode setttings.

Just wondering if this is the expected behavior, or if I was doing something wrong?

If this is still a setting that only works when set in VSCode Settings, then A) the docs could be clearer that those settings are specific to the VSCode settings, and B) it would be rad if it was technically feasible to have that setting picked up from .prettierrc (as it get's more and more of the settings in the repo, rather than in the editor)

Here is what my .prettierrc looks like:

```
{
  "eslintIntegration": true
}
```

CiGit:
This is specific to vscode.
I don't see why it's unclear. It's in a [section](https://github.com/prettier/prettier-vscode#vscode-specific-settings) named **VSCode specific settings**

We could pick it up from there, but what would do prettier (command line) with it? Other editors?
prettier's output:
`[warn] Ignored unknown option: eslintIntegration` is prettier's output.

duro:
The challenge in facing is that in order to get the only if configuration
file present, since eslintIntegration cannot be set there, I have to put a
dummy key in the config any way. You are right that this breaks other
editors. I just hate that if I'm using eslintIntegration I in a sense have
to have my settings in two places in order to use that feature.

CiGit:
Doesn't this work with an empty bag `{}`

duro:
Sorry for the typos in that last message, it was typed fast on a mobile keyboard.

Since I cannot edit on Github mobile, I will put the edit here:

The challenge I'm facing is that in order to get the "only if configuration file is present" setting to work, since eslintIntegration cannot be set there, I have to put a dummy key in the config or duplicate my settings in eslintIntegration and the prettier config. 

You are right that the dummy key breaks other editors. 

I just hate that if I'm using eslintIntegration I have to dupe my settings in two places in order to use the "only if config" feature.

duro:
It does not work with empty bag {}, tried that first

CiGit:
As you have a `.prettierrc` file, putting a prettier option in it isn't that constraining IMHO.

NotMyself:
Hey guys, just stumbled on this thread because I could not seem to get the `prettier.eslintIntegration` setting to work. My confusion was from attempting to set it in my package.json. When I discovered this issue, I realized that it was a setting in vscode directly.

[...]

PR #296, opened 2017-12-12
added text to clarify VSCode settings

NotMyself:
VSCode settings need to be set in VSCode.

Clarifies #292


CiGit:
Thank you for your PR !
May avoid some troubles ;-)

PR #307, opened 2017-12-17
Update Readme for config

CiGit:
Like the cake, this is a lie.

This is not exactly true, but I'd like to keep the Readme simple.


RobinMalfait:
Do you need to add additional changes or can I merge this @CiGit ?

CiGit:
Nothing I can think about right now. If it's clear, it can be merged.

I can still open a new one if needed.

RobinMalfait:
Alright, merged it 😎

PR #310, opened 2017-12-19
Disable formatter based on disableLanguages

### LINKED ISSUE #272: "prettier.disableLanguages" make the vscode self formatter not work.
I wan't to use vscode's self "vscode-typescript" to format the js file.
So i add this:
```js
"prettier.disableLanguages": [
        "javascript"
]
```
Then prettier doesn't work for the javascript file , but, i can't format javascript file now.
It seems that the "vscode-typescript"  has also been disabled.

So does the "json" file.
I need your help  >_<#@!

CiGit:
Status: Ready

TODO:
- [x] Handle untitled files (not persisted)
- [x] Update formatter registration on workspace folder add / remove
- [x] Tests
- [x] ~Certainly some refactor (use a map or 2 `workspaceFolder -> languageSelector`)~ Keep it simple. Recompute everything.
- [x] Changelog
- [x] Readme. It will certainly need a restart required. Or recompute on config change.
- [x] ~Add logging to output panel with enabled language. Should be easier for users to know which language they may disable.~ We should have something like a debug log with more than just that.

fix #272

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 package.json
 package-lock.json
 testProject/
+testWorkspaceFolder/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,7 +19,7 @@
             "request": "launch",
             "runtimeExecutable": "${execPath}",
             "args": [
-                "${workspaceRoot}/testProject",
+                "${workspaceRoot}/testWorkspace.code-workspace",
                 "--extensions-dir=${workspaceRoot}/.", // no other extension loaded
                 "--extensionDevelopmentPath=${workspaceRoot}",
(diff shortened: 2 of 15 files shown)


CiGit:
This works well in case a workspace folder is not a sub folder.
If someone has an idea how to deal with that case. Feel free to help :-)

Doc: https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-nameglobpatternaspan-classcodeitem-id1411globpatternspan

Example case: 
```
wf1                      // js
wf1/folder/shouldFormat.js
wf1/folder/wf2     // no-js
wf1/folder/wf2/shouldntFormat.js
```

CiGit:
I think we could let it as is until a better version. IMO it is an edge case.
Just updated the readme to mention this.

PR #326, opened 2018-01-10
Update Prettier to 1.10.1

### LINKED ISSUE #325: Prettier v1.10 Support
Today's v1.10 release (https://prettier.io/blog/2018/01/10/1.10.0.html) brings support for Vue single-file components and would be great to have inside of VSCode!

felixfbecker:
Was desperately waiting for this for getting formatting on JSON-with-comments (https://github.com/prettier/prettier/pull/3496)

On a sidenote, what about something like Greenkeeper or Renovate to automate this?

Edit @CiGit: fix #325

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3333,9 +3333,9 @@
       "integrity": "sha1-gV7R9uvGWSb4ZbMQwHE7yzMVzks="
     },
     "prettier": {
-      "version": "1.9.2",
-      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.9.2.tgz",
-      "integrity": "sha512-piXx9N2WT8hWb7PBbX1glAuJVIkEyUV9F5fMXFINpZ0x3otVOFKKeGmeuiclFJlP/UrgTckyV606VjH2rNK4bw=="
+      "version": "1.10.2",
+      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.10.2.tgz",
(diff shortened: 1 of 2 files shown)


CiGit:
I would first have #310 in as we would conflict with @octref's [Vetur](https://github.com/vuejs/vetur) I think (I don't use that extension) on formatting vue.

I don't know Renovate and barely used Greenkeeper. Some quick differences?
We may need some more tests tho. Any idea?

felixfbecker:
Sorry, I don't see the relation to #310? Are you specifically talking about json-with-comments?
Also, why does the json-with-comments support not work automatically after updating prettier in a project?

They don't differ much. There are a lot of tools out there nowadays (dependencies.io, dependabot, ...). Renovate has a config file and I think Greenkeeper requires more work to update lockfiles.

Anything specific you would want to test? I was impressed that there are tests at all for a vscode extension

azz:
1.10.2 was just published

CiGit:
>Sorry, I don't see the relation to #310? Are you specifically talking about json-with-comments?

v1.10 support vue files.

>Also, why does the json-with-comments support not work automatically after updating prettier in a project?

We use bundled prettier to define languages.

>They don't differ much. There are a lot of tools out there nowadays (dependencies.io, dependabot, ...). Renovate has a config file and I think Greenkeeper requires more work to update lockfiles.

May be worth trying out.

>Anything specific you would want to test? I was impressed that there are tests at all for a vscode extension

Nothing special in this update.
We started without tests when this extension was basic (it formats or not). :smile:

felixfbecker:
Updated

Issue #332, opened 2018-01-16
How to revert to previous version ( prettier-eslint )


antonreshetov:
How to revert to previous version? Where do you even find all versions of the plugin?

antonreshetov:
Since no one answered, had to ask Google) May be useful to someone link for download the correct version of the plugin

https://${publisher}.gallery.vsassets.io/_apis/public/gallery/publisher/${publisher}/extension/${extension_name}/${version}/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage

Thanks for the article - https://tpodolak.com/blog/2017/04/09/downgrading-visual-studio-code-extension/

azz:
Curious why you had to downgrade?

antonreshetov:
@azz I was faced with a similar problem https://github.com/prettier/prettier-vscode/issues/318

ryanditjia:
The auto-update to 1.1.1 broke prettier-eslint integration for me.

**My environment:**
- macOS (affected my 2 machines)
- VSCode 1.19.2
No global installation, local installation as follows:
![image](https://user-images.githubusercontent.com/16177372/35036446-f67f777c-fba6-11e7-80ce-e8c9b697b742.png)

Downgrading to 1.1.0 fixed it for me.
Here’s how to downgrade if you‘re facing the same problem:
1. Download 1.1.0 from this link
[https://esbenp.gallery.vsassets.io/_apis/public/gallery/publisher/esbenp/extension/prettier-vscode/1.1.0/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage](https://esbenp.gallery.vsassets.io/_apis/public/gallery/publisher/esbenp/extension/prettier-vscode/1.1.0/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage)
2. **Rename** the extension from .VSIXPackage to **.vsix**
3. In your VSCode user settings, add:
`"extensions.autoUpdate": false,`
4. Open up Command Palette (⌘ ⇧ P on macOS) and select **"Extensions: Install from VSIX..."**
5. Select the .vsix file and then reload / reopen VSCode

azz:
This obviously needs further investigation but if upgrading from `1.1.0` to `1.1.1` causes issues, you could try pinning `prettier-eslint` back to `8.3.1` as that's all that happened.

https://github.com/prettier/prettier-vscode/compare/v1.1.0...v1.1.1

[...]

PR #335, opened 2018-01-17
test for eslint integration

CiGit:
Context #332 

- [x] remove test commit (npm tarball) once fixed in dependency

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,5 @@
 package.json
 package-lock.json
+testEslint/
 testProject/
 testWorkspaceFolder/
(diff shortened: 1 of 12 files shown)


CiGit:
Fails as expected

CiGit:
Second commit adds a npm tarball with a patch: prettier/prettier-eslint#161

zimme:
So there was not really a problem with `8.8.0`, however automatically turning off the `prettier/prettier` rule, if found, and still inferring from its rules caused some confusion about the config precedence.

I've reverted that change in `8.8.1` and the behavior should be more as "expected".

CiGit:
I thought about something like this as it was partially respecting the config.

Thanks @zimme , 8.8.1 passes my little test :-)

PR #365, opened 2018-02-06
Prettier silently doesn't work when wrong ignore path provided (Windows 10)

agjs:
error image

When provided with a wrong "prettier.ignorePath", Prettier will simply never get that little tick in the bottom right corner and will completely fail (as shown on the screenshot attached).

As you can see the error throws:
> Cannot read property 'ignores' of undefined: TypeError: Cannot read property 'ignores' of 

This is due to the fact that is trying to read the path that doesn't really exist.

It took me like 2 hours to debug this thanks to the Developer: Startup Performance feature, otherwise, there's no way I would have fixed it.

The simple solution is to simply check if the directory exists and if not, return null.

Better solution for this whatsoever to somehow inform the user that the ignore path he or she used is wrong.

### DIFF
--- a/src/ignoreFileHandler.ts
+++ b/src/ignoreFileHandler.ts
@@ -84,6 +84,9 @@ function getIgnorePathForFile(
     if (!ignorePath) {
         return null;
     }
+    if (!existsSync(ignorePath)) {
+        return null;
+    }
     if (workspace.workspaceFolders) {
         const folder = workspace.getWorkspaceFolder(Uri.file(filePath));
         return folder ? getPath(ignorePath, folder.uri.fsPath) : null;
--- a/src/ignoreFileHandler.ts
+++ b/src/ignoreFileHandler.ts
@@ -86,7 +86,7 @@ function getIgnorePathForFile(
         return null;
     }
     if (!existsSync(ignorePath)) {
-        addToOutput("Wrong prettier.ignorePath provided in your User or Workspace Settings. The path does not exist.");
+        addToOutput(`Wrong prettier.ignorePath provided in your settings. The path (${ignorePath}) does not exist.`);
         return null;
     }
     if (workspace.workspaceFolders) {
--- a/src/ignoreFileHandler.ts
+++ b/src/ignoreFileHandler.ts
@@ -2,6 +2,7 @@ import { readFileSync, existsSync } from 'fs';
 import * as path from 'path';
 import { workspace, Uri, Disposable } from 'vscode';
 import { getConfig } from './utils';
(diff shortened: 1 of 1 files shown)


CiGit:
We should certainly catch that exception and show it in the output panel.

agjs:
@CiGit I agree for sure, that's why I kinda said that would be a better way to do it instead of silently leaving it not doing anything.

CiGit:
Woops, did not noticed this was a PR !

CiGit:
This allows to output a message.
https://github.com/prettier/prettier-vscode/blob/e5d4392d7cf1a85f190bfe3b66e5c49c7baca62a/src/errorHandler.ts#L100

agjs:
Awesome, will add it as part of this PR today and then you can review and merge.

agjs:
Should be fine now, please review, (modify if necessary) and merge.

src/ignoreFileHandler.ts
```diff
@@ -84,6 +85,10 @@ function getIgnorePathForFile(
     if (!ignorePath) {
         return null;
     }
+    if (!existsSync(ignorePath)) {
+        addToOutput("Wrong prettier.ignorePath provided in your User or Workspace Settings. The path does not exist.");
```
CiGit: It also could be from Folder settings. Shorten to settings.
We should also print `ignorePath` saying it doesn't exist.

agjs: Doing it ASAP.

agjs: Updated, please check.

CiGit:
Thanks !

agjs:
Thanks man, will help further on and try to help with the existing reported issues :-). Cheers

CiGit:
Seems wasn't the good fix.
Tests are working on travis but not locally...
I'm fixing it asap

agjs:
Maybe simply tests have to be updated?

CiGit:
Well I don't really understand why test are failing locally as they should, but not on travis...
Something about working directory I think.
Fix is `existSync(absolutPath)`
I've committed it to master directly and will release

CiGit:
published

agjs:
Awesome :-). If you guys ever need more help, making me a contributor would motivate me to help a lot more and fix issues with you guys. Just heads up in case it happens. Cheers

CiGit:
If you'd like, finding why travis passed on this PR but not locally could help a lot!

I see you as a Contributor ;-)

agjs:
Sounds perfect, will take a look at it after the work day ends! :).

PR #378, opened 2018-02-25
Adding initial support for Visual Studio Live Share

### LINKED ISSUE #366: Show prettier.config.js errors
When `prettier.config.js` throws an error, the extension "doesn't work" (files are not fromatted) and the user is not notified: nothing is printed in the **Output** tab, the error is only logged in the developer tools console (which the user probably doesn't open). I don't know if VSCode should be catching those errors and report them in **Output** (in which case I guess this would be a VSCode bug), or if the extension should take care of that. I remember other errors being reported there (eg issue #341), so I thought it might be an issue with the extension.

You can use the following `prettier.config.js` to reproduce the issue:
```js
throw new Error("prettier.config.js error");
```

lostintangent:
This PR adds initial support for [Visual Studio Live Share](http://aka.ms/vsls), which will allow developers who have installed this extension, to continue using Prettier while collaborating on someone else's project (e.g. pair programming). Without this change, if the "guest" in a collaboration session tried to format a range or document, they would get the inbox VS Code formatting behavior, which could be a little confusing, and definitely not as awesome 🚀 

Prettier's `resolveConfig` method throws an exception if it can't enumerate the workspace directories, which in the case of a "guest" in a Live Share session, wouldn't exist, since they've viewing/editing remote files that aren't downloaded locally. 

Long-term, it would be ideal if this extension could use a VS Code-specific API to resolve the app's actual config, using the virtual file system that Visual Studio Live Share exposes (the `vsls:` scheme). With this change, the Live Share "guest" would be able to format, but it wouldn't respect any Prettier config that the app might have. For now though, this PR unblocks the basic scenario of being able to format, for apps that are simply using the default Prettier config.

EDIT @cigit:
 - fix #366

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -43,6 +43,7 @@ async function resolveConfig(filePath: string, options?: { editorconfig?: boolea
     try {
         return await bundledPrettier.resolveConfig(filePath, options);
     } catch (e) {
+        addToOutput(`Failed to resolve config for ${filePath}. Falling back to the default config settings.`);
         return null;
     }
  }
@@ -124,9 +125,10 @@ async function format(
     if (!hasConfig && vscodeConfig.requireConfig) {
         return text;
     }
-    const fileOptions = await resolveConfig(fileName, {
+
+    const fileOptions = (hasConfig ? await resolveConfig(fileName, {
         editorconfig: true,
-    }) || {};
+    }) : null) || {};
 
     const prettierOptions = mergeConfig(hasConfig, customOptions, fileOptions, {
         printWidth: vscodeConfig.printWidth,
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -31,20 +31,23 @@ const STYLE_PARSERS: ParserOption[] = ['postcss', 'css', 'less', 'scss'];
  * @param filePath file's path
  */
 async function hasPrettierConfig(filePath: string) {
-    return (await resolveConfig(filePath)) !== null;
+    const { config } = await resolveConfig(filePath);
+    return config !== null;
 }
+
+type ResolveConfigResult = { config: Partial<PrettierConfig>, error?: Error };
 
 /**
  * Resolves the prettierconfig for the given file.
  * 
  * @param filePath file's path
  */
-async function resolveConfig(filePath: string, options?: { editorconfig?: boolean }): Promise<PrettierConfig | null> {
+async function resolveConfig(filePath: string, options?: { editorconfig?: boolean }): Promise<ResolveConfigResult> {
     try {
-        return await bundledPrettier.resolveConfig(filePath, options);
-    } catch (e) {
-        addToOutput(`Failed to resolve config for ${filePath}. Falling back to the default config settings.`);
-        return null;
+        const config = await bundledPrettier.resolveConfig(filePath, options);
+        return { config };
+    } catch (error) {
+        return { config: {}, error };
     }
  }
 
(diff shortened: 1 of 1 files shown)


CiGit:
Thanks for your PR!

>Prettier's resolveConfig method throws an exception if it can't enumerate the workspace directories, which in the case of a "guest" in a Live Share session, wouldn't exist, since they've viewing/editing remote files that aren't downloaded locally.

This is basically also fixing #366, the root cause is not the same :+1: 
Could you add a message to output panel when it fails to read a config file for whatever reason? `addToOutput(message: string): void`

Concerning long-term, I think this would mean re-implement every part which reads a file:
- prettier config
- prettier ignore
- eslint
- stylelint
- more? (tslint)

What if only the host (You called it _inbox_ if I understood it right) could format? Send a format command to host?
I know really little from this feature. We should open an issue to discuss about that.

src/PrettierEditProvider.ts
```diff
@@ -31,9 +31,22 @@ const STYLE_PARSERS: ParserOption[] = ['postcss', 'css', 'less', 'scss'];
...
 
+/**
+ * Resolves the prettierconfig for the given file.
+ * 
+ * @param filePath file's path
+ */
+async function resolveConfig(filePath: string, options?: { editorconfig?: boolean }): Promise<PrettierConfig | null> {
+    try {
+        return await bundledPrettier.resolveConfig(filePath, options);
+    } catch (e) {
```
CiGit: We shouldn't do that silently
`addToOutput(message: string): void`

lostintangent:
@CiGit Long-term, as you mentioned, I think the best solution would be to "remote" the format action from the guest to the host, such that if the host has this extension installed, then the guest could trigger formatting. Additionally, since the formatting would happen on the host-side, that would naturally pick up any config settings, and therefore, run the correct format. We'll tackle this work and will circle back when we've made progress.

This PR, however, supports the scenario where the host isn't even using Prettier, but the guest might want to. Or, the scenario where the app/host is using Prettier, but with the default settings, and therefore, the guest wouldn't even need access to the config anyways.

src/PrettierEditProvider.ts
lostintangent: Done!

src/PrettierEditProvider.ts
```diff
@@ -124,9 +125,10 @@ async function format(
     if (!hasConfig && vscodeConfig.requireConfig) {
         return text;
     }
-    const fileOptions = await resolveConfig(fileName, {
+
+    const fileOptions = (hasConfig ? await resolveConfig(fileName, {
```
CiGit: `hasConfig` means there is a prettier config (without editorconfig)
We still need to read editorconfig

In this case this means printing the error message twice a run.
I think you wrote this ternary to correct this 😄 

Our `resolveConfig` (the one you just wrote) could return
```js
{config: PrettierConfig|null, error?: Error}
```
and log only later.
I'm opened to better ideas!

CiGit:
Seems you have your hands on both projects 😃

src/PrettierEditProvider.ts
lostintangent: Yeah I was trying to work around the double-logging issue with the ternary, which is very hack-ish, but I was trying to be conservative about the changes I made :) I'll make the change you proposed and will updated the PR. Thanks!

lostintangent:
@CiGit I'm on the Live Share team within Visual Studio, and am also a huge Prettier fan :)

lostintangent:
@CiGit OK, I think this solution is a lot cleaner (thanks for your feedback!) and works beautifully. Let me know if you have other thoughts.

src/PrettierEditProvider.ts
```diff
@@ -31,9 +31,26 @@ const STYLE_PARSERS: ParserOption[] = ['postcss', 'css', 'less', 'scss'];
  * @param filePath file's path
  */
 async function hasPrettierConfig(filePath: string) {
-    return (await bundledPrettier.resolveConfig(filePath)) !== null;
+    const { config } = await resolveConfig(filePath);
+    return config !== null;
 }
 
+type ResolveConfigResult = { config: Partial<PrettierConfig>, error?: Error };
```
lostintangent: Happy to "inline" this type and/or move it to another file. This seemed like a simple solution to start though.

CiGit: This type is local to this file, it fits well in there.
Could be an inline type as it is used once ... But it is also good as it is :+1:

CiGit:
I find it clean! Thanks a lot for your work ! :tada:

src/PrettierEditProvider.ts
CiGit: I also have on my todo list:
Clean/correct our types

PR #416, opened 2018-04-13
[Discussion] [Live Share] Restricting workspace formatting to local files

lostintangent:
As a continuation of #378, this PR simply updates the existing workspace `DocumentSelector`, so that it only applies to `file:`-based, and `untitled` (unsaved) documents. In the case of Visual Studio Live Share, a "guest" would see documents using the `vsls:` scheme, and would therefore "remote" the formatting request to the "host" as opposed to trying to run Prettier locally. However, this change wouldn't impact the "traditional" use of Prettier.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -49,6 +49,7 @@ function selectorsCreator(wf: WorkspaceFolder) {
                 ({
                     language: l,
                     pattern: relativePattern,
+                    scheme: 'file'
                 } as DocumentFilter)
         );
     }


CiGit:
You mean this is enough to enable _remote_ formatting ? :tada: 

Could there be any other scheme were it would break this extension ?

lostintangent:
Yep! The other main scheme is “untitled”, which represenrs unsaved files. But I wasn’t sure if you wanted to support that or not, so I just made this simple change to start.

lostintangent:
@CiGit Any other thoughts about this proposal? It's fully backwards compatible with the existing extension, and so shouldn't introduce any impacts on the user experience.

src/extension.ts
```diff
@@ -49,6 +49,7 @@ function selectorsCreator(wf: WorkspaceFolder) {
                 ({
                     language: l,
                     pattern: relativePattern,
+                    scheme: 'file'
```
RobinMalfait: You know when doing code reviews that the following situation happens more than often: When there are very few changes, like in this PR, then the code review is done with high precision and you will get more comments than that you actually wrote lines of code.

So to fix this I am going to write this useless comment to prove that my above statement is true and that I can comment way too much information for a little line change, which looks perfect by the way!

Also, when a PR is has many many files changed and is fairly big then the code reviews sometimes look like:

> LGTM

But not this one, this PR is small and thus needs a long comment. Okay okay, I frankly don't know what I am doing at this point but I hope you are enjoying your day today. 🎉

CiGit:
Nothing more :-) 
Thanks!

Issue #254, opened 2017-11-05
Setting to specify what .prettierrc to use


felixfbecker:
We share a .prettierrc between multiple repos by publishing it as an npm module, running `prettier` with
```
prettier --config ./node_modules/@sourcegraph/prettierc/.prettierrc
```
prettier-vscode currently does not have a way to configure that, it will not detect any config for the repo, so we have to duplicate the settings, which defeats the purpose of sharing the config. It would be nice to have a setting `"prettier.configPath"` like `"prettier.ignorePath"`. I know you don't like new settings but this one is to avoid having to use the other settings ;)

azz:
Workaround, create a `.prettierrc` with:

```js
module.exports = require('@sourcegraph/prettierc');
```

felixfbecker:
That's a great idea!

kutyel:
Regarding this, according to prettier's docs, the `.prettierrc` file could also be  a .yml file but when I try to do so the JSON validation yells at me. Is this intentional? Should I open another issue for this? Thanks in advance! <3

azz:
If you're working with yaml in vscode I'd recommend using `.prettierrc.yml` instead.

CiGit:
@kutyel our extension is associating .prettierrc to JSON. You can change it to yaml if you prefer and it won't yell. Maybe we should remove that default association. And validate only .prettierrc.json

kutyel:
I don't know what is the correct behaviour (cause I am biased) but for me is kinda counter-intuitive that *prettier* itself allows those formats (without changing the extension of the file) and the `vscode-extension` itself doesn't. BTW the apart from that it works perfectly and thanks very much guys for creating it for VSCode! ❤️

kachkaev:
@felixfbecker if you prettier config is in `./node_modules/@sourcegraph/prettierc/.prettierrc`, can you simply create a symlink at the root of your project pointing to that file? That might fix the issue you are having, unless I've misunderstood it.

felixfbecker:
I like to avoid symlinks if they are not needed. Many tools do not follow them (because there is risk of infinite loops) and they don’t work by default on Windows <10 / need special permissions.

ntotten:
Feature added in version 3.0

PR #433, opened 2018-05-01
Add prettier.config setting

gakada:
For https://github.com/prettier/prettier-vscode/issues/254.


gakada:
Did some changes. `getAbsolutePath` from `utils` is used for path handling for both ignore and config files. For `configCacheHandler`
```ts
const prettierConfigPath = getConfig().config;

if (prettierConfigPath) {
    PRETTIER_CONFIG_FILES.push(basename(prettierConfigPath));
}
```
the intention is to watch for prettier config files, for setting files restart is required, same as for `ignorePath` setting (noted in readme).

CiGit:
Sorry for the delay.

Quickly went through (I will need a little more time to review), seems fine. 
Last question (hopefully), there is a check `hasPrettierConfig`, shouldn't it return true if a config path is given **and** that config file exists ?

gakada:
Right, so for files like `.prettierrc` default configuration is (mostly) ignored (so it is `bundledPrettier` configuration). Added a commit for the same behavior when `config` is set.

Another thing, if `config` is set but doesn't exist, regular config files like `.prettierrc` are ignored, but that is also how [`prettier.resolveConfig`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath-options) is.

tvald:
I'd love to see this feature merged.
@gakada any chance to fix the conflict? (I'm happy to fix and make a new PR)
@CiGit was your question addressed above?

src/configCacheHandler.ts
```diff
@@ -15,6 +17,12 @@ const PRETTIER_CONFIG_FILES = [
     'prettier.config.js',
 ];
 
+const prettierConfigPath = getConfig().config;
```
CiGit: There may be multiple config files, one per workspace folder.
This is just reading one.

To do it correctly, we would need to set up file change hooks for every files. And remove them when needed. 

When a workspace folder is added, check the config and register a hook for it if needed
When a workspace folder is removed, remove the file watcher if it exists

src/configCacheHandler.ts
```diff
@@ -15,6 +17,12 @@ const PRETTIER_CONFIG_FILES = [
...
+
+if (prettierConfigPath) {
+    PRETTIER_CONFIG_FILES.push(basename(prettierConfigPath));
```
CiGit: It shouldn't be in this array, see previous comment.

src/utils.ts
```diff
@@ -54,3 +55,21 @@ export function getGroup(group: string): PrettierSupportInfo['languages'] {
 function getSupportLanguages(version?: string) {
     return (require('prettier') as Prettier).getSupportInfo(version).languages;
 }
+
+/**
+ * Get absolute path for relative or absolute filePath.
+ * Relative to uri's workspace folder in relative case.
+ */
+export function getAbsolutePath(uri: Uri, filePath: string): string | undefined {
```
CiGit: :+1:

ntotten:
Closing as this is very outdated.

Issue #393, opened 2018-03-15
Usability: error feedback easy to miss


joliss:
When an error occurs during prettification, it can look like Prettier is just failing silently, with no apparent cause.

For example, say I have the following JavaScript file:

```js
export { foo };

export function foo() {
  console.log(       1,    2,          3);;
}
```

When I run the "Format Document" command, I'm expecting the file to be reformatted. However, since there is a syntax error that I didn't notice (`foo` is exported twice), prettification fails. To me, it looks like nothing is happening at all; the file is simply not being formatted.

After some sleuthing, I discovered that there is a clickable "Prettier" entry in the status bar, and that it changes to "X" when an error occurs:

<img width="484" alt="status bar" src="https://user-images.githubusercontent.com/524783/37483500-0dff7c90-287e-11e8-839c-d3772be0d52a.png">

It was not obvious to me that I should look for an error indicator in the status bar. For the longest time, I assumed that "Format Document" would simply silently stop working at random times, and I wasn't sure why.

Perhaps usability could be improved by providing a clearer error notification.

RobinMalfait:
Since the notifications are now in the bottom left we have to re add them? Are you willing to provide a PR?

joliss:
There's a bunch of other open source things I need to do, so I'm afraid it's unlikely I'll be able to send a PR for this in the near future.

RobinMalfait:
I'll see if I can find some time in the near future. Thanks for the report though! Once I start working on it I'll mention you so that we can ping pong some ideas to make it usable!

atombender:
As someone who does not have the status bar turned on, this is a pretty annoying problem. Prettier frequently stops working for unknown reasons. ESLint has the same issue, where operational errors (e.g. a package referenced by `.eslintrc` doesn't load) are hidden in the "ESLint Output" panel.

RobinMalfait:
Ok I am going to bookmark this and add it to my todos since I totally forgot about this...

CiGit:
Back with notifications !

etiennejcharles:
@CiGit @joliss 

Is there something that pushed us to revert that commit? 
It was merged, but reverted 2 days right after [here](https://github.com/prettier/prettier-vscode/commit/ce62204e52b41064b195d9346cd588bea264eb46). 

And yup I did expect some visual feedback more _violent_  than the "X" I looked for maybe an hour scratching my head, to find out it wasn't my HTML formatter not working, but rather "prettier" being the issue - 

[...]

PR #434, opened 2018-05-01
add an error message popup when prettier fails to format

RobinMalfait:
This PR will add an error popup when prettier fails to format. In earlier version we also had this feature but the popups were at the top and very intrusive.

Now with the new notifications they appear in a corner and are not taking over focus. When you press the <kbd>esc</kbd> key, they are dismissed.

The notifications also has a `Show error` button which will open the output panel when triggered. This will show the full error.

Note: I tried putting the error inside the notification directly, but it removed all new lines resulting in the output code to be a one liner, unreadable.

Feel free to discuss options or try this feature out!

It will solve the following issues: #393

An example of the popup:

### DIFF
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -15,9 +15,21 @@ let statusBarItem: StatusBarItem;
 let outputChannel: OutputChannel;
 let prettierInformation: string;
 
+function showPrettierErrorMessage() {
+    const showErrorAction = 'Show error';
+
+    window
+        .showErrorMessage('Prettier failed to format!', showErrorAction)
+        .then(action => {
+            if (action === showErrorAction) {
+                commands.executeCommand('prettier.open-output');
+            }
+        });
+}
+
 function toggleStatusBarItem(editor: TextEditor | undefined): void {
     if (editor !== undefined) {
-        // The function will be triggered everytime the active "editor" instance changes
+        // The function will be triggered every time the active "editor" instance changes
         // It also triggers when we focus on the output panel or on the debug panel
         // Both are seen as an "editor".
         // The following check will ignore such panels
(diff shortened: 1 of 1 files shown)


CiGit:
Wouldn't this be a chance for us to review all error things. We have user input in different place which can go wrong.

src/errorHandler.ts
```diff
@@ -15,9 +15,21 @@ let statusBarItem: StatusBarItem;
 let outputChannel: OutputChannel;
 let prettierInformation: string;
 
+function showPrettierErrorMessage() {
+    const showErrorAction = 'Show error';
+
+    window
+        .showErrorMessage('Prettier failed to format!', showErrorAction)
```
CiGit: What about putting an other message in there ? Parameter? I see us use it a little more in the future if this new notification center fits out needs.

Quick idea: `addToOuput` shows a notification with a short message. `More` button to open the panel with the complete error message ? What do you think?

src/errorHandler.ts
```diff
@@ -15,16 +15,25 @@ let statusBarItem: StatusBarItem;
...
-        .then(action => {
-            if (action === showErrorAction) {
-                commands.executeCommand('prettier.open-output');
-            }
-        });
+function showPrettierErrorMessage(message: string = 'failed to format!') {
+    const showErrorAction = 'More';
+
+    // Prefix the message with Prettier so that the user knows which extension it is
+    const fullMessage = `Prettier: ${message}`;
```
CiGit: From a screenshot you provided, It seems VSCode already adds _Source: Prettier_

rudyhuynh:
Is there a setting to suppress the popup?

RobinMalfait:
@rudyhuynh nope

PR #442, opened 2018-05-06
Prefer local `prettier-eslint` over bundled version

dtinth:
The bundled version of `prettier-eslint` [does not support `.mjs` files](https://github.com/prettier/prettier-eslint/pull/173), so I had to use my fork. However, this extension uses the bundled version of `prettier-eslint` instead of the local version installed in the project.

This PR fixes it. :)

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -154,7 +154,7 @@ async function format(
     if (vscodeConfig.eslintIntegration && doesParserSupportEslint) {
         return safeExecution(
             () => {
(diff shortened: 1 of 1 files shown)


CiGit:
I see this as potentially more issues. Context: https://github.com/prettier/prettier-vscode/pull/92#issuecomment-296450437

We would happily update prettier-eslint when (if?) your PR makes it in prettier-eslint.

If you really need it, you can install your fork in the extension folder.

dtinth:
Thanks for the information :)

PR #444, opened 2018-05-07
adds warning about popular formatting packages

capaj:


capaj:
I'll try to rephrase.

README.md
```diff
@@ -14,6 +14,8 @@ Can also be installed using
 ext install prettier-vscode
 ```
 
+A word of warning-if you have any other code formatting extensions installed such as for example hugely popular `HookyQR.beautify` or `taichi.react-beautify` they might take precedence and format your code instead of prettier leading to unexpected results.
```
lipis: `prettier` -> `Prettier`

README.md
```diff
@@ -14,6 +14,8 @@ Can also be installed using
 ext install prettier-vscode
 ```
 
+A word of warning-if you have any other code formatting extensions installed such as for example hugely popular `HookyQR.beautify` or `taichi.react-beautify` they might take precedence and format your code instead of Prettier leading to unexpected results.
```
RobinMalfait: I think that the message itself is good, but it might be a good idea to get the attention of the developer because now it is just a piece of text. If we prepend the message with something like:

```
> ⚠️ A word of warning, if ...
```

RobinMalfait:
General question, should we mention some workarounds as described here? https://github.com/prettier/prettier-vscode/issues/454#issuecomment-389311110

capaj:
@RobinMalfait that's only workaround for jsbeautify. There are many other formatting extensions which are in conflict. I would not bother with it.

RobinMalfait:
@capaj that's true. Once @CiGit approves this we can merge it.

RobinMalfait:
@CiGit Only 1 way to find out :D

Issue #481, opened 2018-06-04
Disable error messages.


noobiept:
Is there a way to disable the messages that appear at the bottom-right? I find it extremely distracting, since they appear while I'm still writing the code and don't go away either.

CiGit:
It appears only when you format something and there is an error. You can press <kbd>ESC</kbd> to hide it.

noobiept:
Any way to disable them?

RonanQuigley:
Just to second this: the inline-error on its own is obvious enough for me. Having pop-ups as well gets really annoying.

Using escape works, but it should just have an option to turn it off. . Your only option seems to be to disable:

``` 
"[javascript]": {
        "editor.formatOnSave": false
    },
```

though that obviously means format on save is gone. Is this a new feature as I can't recall having this as a problem before?

CiGit:
Well, thinking about it.
This is really meant to attract user's attention: Something went wrong.
In most cases, these are syntax errors. Seems redundant with a linter. Some users seem happy with it.

In my case, I don't need a notification for a syntax error. But I would need it for any other uncommon error this extension may produce (reading fs mostly). But as I said, those notification seem to please some users.

noobiept:
There's already plenty of clues about the issue being presented on the editor, to me an extra message is redundant, but again, thats my opinion, other people might prefer to have the extra message. 

Thats solvable by adding a setting, this way people can make it work the best for them.

RonanQuigley:
Yep @noobiept sums it up. Furthermore, if like me, you use a lot of unit testing. you already get continuous feedback for this sort of thing. I can see its use, but this is just bloat in my setup.

starquake:
Agree with @noobiept. It distracts from all the other things that are also errors.

dnutels:
This used to work though... (as in, before the latest updates on VSCode Insiders) there is a status-bar with corresponding icon for a reason.

breathe:
I resoundingly agree with those who don't want the notifications -- especially the notifications don't appear when the syntax errors are corrected ...

These popups are horribly annoying.

kristian-puccio:
Another thing to note is that if you use the vim plugin hitting esc doesn't dismiss the error popup.
So the only way to dismiss it is to use the mouse, which sort of ruins the experience of using vim bindings.

vbarrancos:
[...]

PR #484, opened 2018-06-06
Add an option to toggle error messages off

alexzaworski:
This adds a setting to disable the error messages introduced by https://github.com/prettier/prettier-vscode/commit/ce62204e52b41064b195d9346cd588bea264eb46, closing https://github.com/prettier/prettier-vscode/issues/481.

I've neither written a line of typescript in my life nor worked on a vscode extension so if there's a more idiomatic way to do this (or I've broken something) I'd be happy to make changes, just need some guidance— many of us find these popups pretty frustrating 😢

I suspect I'm supposed to be passing a uri into `getConfig` but could not figure out what that would be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,9 @@ Require a 'prettierconfig' to format
 Supply the path to an ignore file such as `.gitignore` or `.prettierignore`.
 Files which match will not be formatted. Set to `null` to not read ignore files. Restart required.
 
+#### prettier.showErrorMessages (default: true)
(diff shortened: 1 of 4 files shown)


CiGit:
Thanks for your PR!
Sadly I will close that one as we have simply removed the notification message. Until we reintroduce them differently ;-)

alexzaworski:
Works for me :) Thanks for the followup

PR #485, opened 2018-06-06
Parser inference also based on filename and version

### LINKED ISSUE #483: Couldn't resolve parser "json-stringify"
When trying to format json files I am getting this error. 

This happens in our monorepo when attempting to format any JSON file. Repo is here: https://github.com/forcedotcom/salesforcedx-vscode

Prettier VSCode Version 1.4.0
VS Code Version: 1.23.1
Node Version: v8.11.2
NPM Version: 6.0.1

CiGit:
Infer parser by taking into account version and filename.

Change tests, let prettier infer the parser based on filename.
Fix #483

### DIFF
--- a/src/PrettierEditProvider.ts
+++ b/src/PrettierEditProvider.ts
@@ -35,21 +35,24 @@ async function hasPrettierConfig(filePath: string) {
     return config !== null;
 }
 
-type ResolveConfigResult = { config: PrettierConfig | null, error?: Error };
+type ResolveConfigResult = { config: PrettierConfig | null; error?: Error };
 
 /**
  * Resolves the prettierconfig for the given file.
- * 
+ *
  * @param filePath file's path
  */
-async function resolveConfig(filePath: string, options?: { editorconfig?: boolean }): Promise<ResolveConfigResult> {
+async function resolveConfig(
+    filePath: string,
+    options?: { editorconfig?: boolean }
+): Promise<ResolveConfigResult> {
     try {
         const config = await bundledPrettier.resolveConfig(filePath, options);
         return { config };
     } catch (error) {
         return { config: null, error };
     }
- }
+}
 
 /**
  * Define which config should be used.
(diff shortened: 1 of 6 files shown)


src/PrettierEditProvider.ts
```diff
@@ -35,21 +35,24 @@ async function hasPrettierConfig(filePath: string) {
     return config !== null;
 }
 
-type ResolveConfigResult = { config: PrettierConfig | null, error?: Error };
+type ResolveConfigResult = { config: PrettierConfig | null; error?: Error };
```
RobinMalfait: Is this change correct? If so, than I should do some more research on typescript.

src/types.d.ts
```diff
@@ -7,6 +7,7 @@ export type ParserOption =
     | 'scss'
     | 'typescript'
     | 'json'
+    | 'json-stringify'
```
RobinMalfait: That makes sense 👍

src/PrettierEditProvider.ts
CiGit: Prettier did it.
interface or object type can separate their properties with `,` `;` or `\n`

RobinMalfait: Interesting!

src/utils.ts
```diff
@@ -12,10 +13,17 @@ export function getConfig(uri?: Uri): PrettierVSCodeConfig {
...
-    const language = getSupportLanguages().find(lang =>
-        lang.vscodeLanguageIds.includes(languageId)
+    const language = getSupportLanguages(version).find(
+        lang =>
+            lang.vscodeLanguageIds.includes(languageId) &&
+            // Only for some specific filenames
+            (lang.extensions.length > 0 ||
+                (path != null &&
+                    lang.filenames != null &&
+                    lang.filenames.includes(basename(path))))
```
CiGit: Note: this works because 'json-stringify' is before 'json' in the array.

CiGit: Should do better...

PR #550, opened 2018-08-12
Add format to mac os touch bar

casperstr:


CiGit:
Shouldn't this be an extension by itself? I see nothing specific to prettier, only a command: format document
It would work for every formatting extension.

PR #559, opened 2018-08-20
Add 'mongo' as supported language

trungtin:
As the title said. We normally write mongo query like either javascript or json, so prettier will parse and format just right.
Great for formatting cosmos scrapbook too

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -47,6 +47,7 @@ export function rangeSupportedLanguages(): string[] {
         'typescriptreact',
         'json',
         'graphql',
+        'mongo',
     ];
 }
 


CiGit:
I doubt this does anything. Or crashes at worst.
Prettier has no knowledge of `mongo` as a languageId.
And why adding it to the range formatter ?

trungtin:
Sorry for my ignorance, have not really dived into the project to know where to add the language.
When I add in range formatter then vscode will allow me to run `format` on mongo language.

And for example this syntax is totally formattable by prettier, isn't it?
 ```
db.inventory.insertMany([
   { item: "journal", qty: 25, size: { h: 14, w: 21, uom: "cm" }, status: "A" },
   { item: "notebook", qty: 50, size: { h: 8.5, w: 11, uom: "in" }, status: "A" },
   { item: "paper", qty: 100, size: { h: 8.5, w: 11, uom: "in" }, status: "D" },
   { item: "planner", qty: 75, size: { h: 22.85, w: 30, uom: "cm" }, status: "D" },
   { item: "postcard", qty: 45, size: { h: 10, w: 15.25, uom: "cm" }, status: "A" }
]);
```

CiGit:
No worries, I was just trying to understand what you were trying to do.

The range formatter is just a white-list of languageId already known by prettier. (It does not test if the language exists in prettier, being hard coded)
Prettier is able to format javascript, but is unable to map `mongo` to `javascript`.
And if I have it correctly, a mango file can also contain `json`. I don't see how we could support automatically multiple language.

Currently, if you want to format your snippet, you can switch document's language from `mongo` to `javascript` -> Format -> switch language back

CiGit:
VSCode Languages are provided by prettier. If this were a thing, it should be added in there.

Issue #522, opened 2018-07-23
Use global `prettier` package as a fallback when Prettier is not installed inside a project.


ryanpcmcquen:
When inside a directory with a `package.json` that has `prettier` as a dependency, this extension will not format a document until I run `yarn install` (or `npm install`), even if I have `prettier` installed globally on the machine. Seems like the extension should check for a global fallback rather than just failing.

kurtextrem:
This extension has a bundled version integrated, however I agree it should use this method of checking:
`global > local > bundled`

CiGit:
It's true, it fails if you have a prettier declared in your deps but not installed.
It uses a bundled version if it's not specified. 

We should show an error message in the first case.

RobinMalfait:
The order that @yuhr provided makes more sense. If you work in a team and have a certain version locally than you want to use that version. Otherwise use the global one, if you don't have it, use the bundled one.

PR #561, opened 2018-08-21
use bundled prettier if local prettier does not resolve

anisjonischkeit:
This PR is a fix for this issue: https://github.com/prettier/prettier-vscode/issues/522.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 ## [Unreleased]
 
 -   Validate the `"prettier"` key in `package.json` using the prettier settings schema
+-   Prettier [1.15](https://prettier.io/blog/2018/11/07/1.15.0.html)
+-   New options: jsxSingleQuote, htmlWhitespaceSensitivity and endOfLine (More info in readme).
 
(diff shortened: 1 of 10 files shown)


CiGit:
Thanks for your contribution.
I'm not convinced this is the right solution: silently using an other prettier version.

IMHO: If a user specifies a prettier version, they would expect their code be formatted by that given version.

anisjonischkeit:
I didn’t realise you could even specify what version of prettier you want to use. How do you set it?

CiGit:
package.json's (dev)dependencies

anisjonischkeit:
The problem we are having is that our node_modules folder is not in the default location. With the current version of prettier-vscode, it fails silently in this scenario. With the PR, it passes silently. What if we let it pass but display a message saying "Could not find locally installed prettier, maybe try running npm install. Defaulting to bundled prettier."

xeoneux:
Ah! There's already a PR for this! :)

jgoux:
What can we do to move this PR forward?

I'm using yarn `plug and play` feature and there is no node_modules directory, so no local prettier resolved. 😢

CiGit:
Closing in favor of #811

PR #600, opened 2018-10-13
Use the schema for package.json "prettier" key

DJMcNab:
This uses the schema located at <http://json.schemastore.org/prettierrc> for the `package.json` `"prettier"` key, so that it can be autocompleted and validated.

The only unresolved question I have is whether the schema file should be in a different folder or have a different name. I think that the new file will be included in the built extension, as the `.vscodeignore` file does not include it, although I'm not certain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 ## [Unreleased]
 
+-   Validate the `"prettier"` key in `package.json` using the prettier settings schema
+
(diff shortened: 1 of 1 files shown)


svipben:
You shouldn't add yourseld to the contributor's list, @prettier decides not you, just saying.

DJMcNab:
Fixed!

Thanks and sorry @svipben.

DJMcNab:
No problem. One of my biggest pain points with prettier was using the package.json settings, which I think will be better now

PR #619, opened 2018-11-07
feat: update prettier to 1.15.1 and add html test

n1ru4l:


CiGit:
Thanks for your PR.
1.15 also brings in more settings :disappointed: 
- https://prettier.io/docs/en/options.html#jsx-quotes
- https://prettier.io/docs/en/options.html#html-whitespace-sensitivity
- https://prettier.io/docs/en/options.html#end-of-line

Except `EndOfLine` --  which is already in VSCode `files.eol` -- I think we should add them.

n1ru4l:
@CiGit I am on it!

n1ru4l:
@CiGit "files.eol" differs from prettier endOfLine (https://github.com/Microsoft/vscode/blob/409e61c5cadcbd0347b23e3102ae3ebb5258b8f9/src/vs/workbench/parts/files/electron-browser/files.contribution.ts#L239) vs https://prettier.io/docs/en/options.html#end-of-line

So I added it as a prettier option instead of using the vscode option.

n1ru4l:
Also rebased master, this should be ready to merge!

CiGit:
So `\r` is the only difference from what I see. What I meant was not using `file.eol` to pass it to prettier but allow VSCode to change it with it's own tools if needed. Prettier will then work in auto mode.
And if a project really needs something special, we still read it from config files.

Just sharing my thought. I've no strong preference.

n1ru4l:
Makes sense to me, should I remove the `prettier.endOfLine` setting?

CiGit:
It doesn't cost us to keep it. In auto mode it does nothing. Let's keep it

Can you add those new settings to the readme ?

And we are good!

svipben:
Also, don't forget about changelog 😄

n1ru4l:
@CiGit @svipben Done.

README.md
```diff
@@ -120,6 +120,16 @@ Files which match will not be formatted. Set to `null` to not read ignore files.
...
 *Note: Disabling a language enabled in a parent folder will prevent formatting instead of letting any other formatter to run*
 
+#### prettier.jsxSingleQuote (default: false)
+Use single quotes instead of double quotes in JSX.
+
+#### prettier.htmlWhitespaceSensitivity (default: 'css')
+Specify the global whitespace sensitivity for HTML files. [Learn more here](https://prettier.io/docs/en/options.html#html-whitespace-sensitivity)
+
+
+#### prettier.htmlWhitespaceSensitivity (default: 'auto')
```
svipben: I guess this is a typo `htmlWhitespaceSensitivity` and it should be a `prettier.endOfLine`

README.md
```diff
@@ -120,6 +120,16 @@ Files which match will not be formatted. Set to `null` to not read ignore files.
...
+Specify the end of line used by prettier. [Learn more here](https://prettier.io/docs/en/options.html#end-of-line)
+
```
CiGit: Settings shoulb be put in prettier's settings, not vscode settings for prettier.
After `#### prettier.arrowParens (default: 'avoid')`

n1ru4l: Done.

CiGit:
Great! Thank you!

philipithomas:
Thanks all contributors for working quickly, testing the code, and cutting a new release!

Issue #496, opened 2018-06-14
prettier doesn't use eslintConfig when modules are installed globally


coolow:
i have installed prettier and eslint extensions in vscode, and enable prettier.eslintIntegration setting.
and i have installed eslint and eslint-config-airbnb-base modules **globally** in NPM.
but prettier used prettierOptions  rather than eslintConfig when i format codes.
example codes:
![image](https://user-images.githubusercontent.com/302606/41411231-a924e160-700e-11e8-826a-39b82544555b.png)
after formating:
![image](https://user-images.githubusercontent.com/302606/41411257-b6bf8a14-700e-11e8-93b4-fdda3c6002be.png)
however when i installed prettier-eslint-cli module **globally** in NPM. then run `prettier-eslint .\src\index.js`. it worked as expected. see below.
![image](https://user-images.githubusercontent.com/302606/41411356-1f593804-700f-11e8-9509-d14bb6619c3c.png)
i think there are some issues when prettier resolving eslintConfig if modules are installed globally. hope it be fixed.

CiGit:
We do not use your globally installed prettier-eslint and thus can't access any of your globally installed dependencies. We do however pick up your locally installed dependencies. (prettier-eslint does it)

ndstephens:
Unless I missed it in the docs, I didn't see anywhere that in order for _prettier.eslintIntegration:true_ to work you need to have all your eslint dependencies installed locally, not globally.  It makes sense now, and I understand why they should be installed locally per project instead of globally.  However when I was trying to get this to work for the first time, and happened to have everything installed globally, I've been struggling to make it work for a long time...until I came across this issue post.  

Maybe in your docs where you write:

> **prettier.eslintIntegration (default: false) - JavaScript and TypeScript only**
> Use _prettier-eslint_ instead of _prettier_. Other settings will only be fallbacks in case they could not be inferred from ESLint rules.

you could add something like:

> prettier-eslint already installed with extension.  All other eslint dependencies need to be installed locally

Again, maybe i just missed that info, or the assumption that a dev will know to install all eslint dependencies locally is so strong that you feel it doesn't need to be stated (i'm still a bit of a newb).

And to be clear...I say none of this mockingly or sarcastically.  I love the extension and all the hard work that has gone into this.  Just offering up some unsolicited advice from a newish dev.

CiGit:
@ndstephens Would you like to send a PR to clarify that part?

Dudeonyx:
[...]

PR #629, opened 2018-11-09
Update README.md, ESLint integration info

ndstephens:
Provides more detail about the ESLint (and TSLint) integration setup requirements to function properly in VSCode.

Attempting to resolve issue #496 

Not sure what I wrote is entirely correct 😕 , please let me know if adjustments need to be made.


CiGit:
Thanks! It is not entirely correct but being correct would be difficult to explain :smile: 
That's a safe explanation.

Issue #631, opened 2018-11-10
Upgrade to prettier 1.15.2


markgoho:
Update dependency to 1.15.2

CiGit:
Fixed by c6b3a63c638069aea2f8c1e4dcf44d1a5e00e593

PR #632, opened 2018-11-10
Update prettier dependency

markgoho:
Fix for #631


CiGit:
Thanks, sadly this is incomplete. Missing `package-lock.json`.

Master is already updated. I'll close that one.

Issue #647, opened 2018-11-20
If prettier is installed locally as a nested dependency, this extension won't find it


mbrowne:
Currently this extension checks for a locally installed version of prettier but it only finds it if it's explicitly mentioned in the `dependencies` or `devDependencies` of the `package.json` for the project. I've created a fork that allows it to work for nested dependencies too.

Example use case: at my company, we have a separate code-quality package with all of our linting/formatting configuration, and prettier is a dependency of the code-quality package. With the current behavior, our pre-commit can end up running a different version of prettier than the version used by this extension (this happened today because our local prettier version was slightly behind the version bundled with this extension).

mbrowne:
PR: https://github.com/prettier/prettier-vscode/pull/648

CiGit:
This is something we removed. 
#67

It is easier to add prettier as a dependency instead of removing it from a dependency.
Seems your issue could be fixed once we allow to specify a path to prettier.

mbrowne:
Ideally we would like the version of prettier to be determined by the shared code-quality package. But if it were possible to specify a path to prettier, that would be a good solution as long as it's possible to do so on a per-project level.

ntotten:
Duplicate of #232

PR #648, opened 2018-11-20
Added support for prettier installed locally as a nested dependency

mbrowne:
Fixes https://github.com/prettier/prettier-vscode/issues/647.

This PR resolves the issue by avoiding the explicit checking of `dependencies` and `devDependencies`. Instead, it goes ahead and tries to resolve the prettier module regardless, so it will find it if it's a nested dependency.


ntotten:
Closing as this is too out of date.

PR #649, opened 2018-11-21
Update Travis Config

sibiraj-s:
Hi,
Travis will be undergoing an infra upgrade and hence it is advised to remove `sudo: false` from travis-ci config. You can read more here https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: node_js
 node_js:
  - "8"


CiGit:
Thanks for the info. Will need to update some project ASAP :smile:
We do not need to test different node versions. vscode comes with its own version

sibiraj-s:
So I will remove nodejs `8 and 10` and add `statble`. is that okay?

PR #655, opened 2018-11-23
Hide status bar item if no editor is open

sibiraj-s:
this hides status bar item if no editor is opened.

### DIFF
--- a/src/errorHandler.ts
+++ b/src/errorHandler.ts
@@ -47,6 +47,11 @@ export function registerDisposables(): Disposable[] {
                 toggleStatusBarItem(editor);
             }
         }),
+        window.onDidChangeActiveTextEditor(editor => {
+            if (editor === undefined) {
+                statusBarItem.hide();
+            }
+        }),
     ];
 }
 


src/errorHandler.ts
```diff
@@ -47,6 +47,11 @@ export function registerDisposables(): Disposable[] {
                 toggleStatusBarItem(editor);
             }
         }),
+        window.onDidChangeActiveTextEditor(editor => {
```
CiGit: This logic should go into `toggleStatusBarItem(editor)`

sibiraj-s:
Updated @CiGit

src/errorHandler.ts
sibiraj-s: Updated.

sibiraj-s: Updated 👍

sibiraj-s: I didn't notice at all the handler's are same at first. 😅

sibiraj-s:
ping @CiGit

PR #692, opened 2019-01-13
Added Shortcuts Section

umairx97:
Added shortcuts for both Mac and Windows users for those new to VS Code


ntotten:
Solved with 0b07b1f5af83a931f4b360f5386a4ea544e01f13

Issue #719, opened 2019-02-11
Curly braces in HTML files getting weird formatting


trainiac:
In a Vue SSR application you can interpolate raw strings likes this
```html
<!DOCTYPE html>
<html>
  <head>
    {{{ head }}}
    {{{ renderResourceHints() }}}
    {{{ renderStyles() }}}
  </head>
  <body>
    {{{ renderState() }}}
    {{{ renderScripts() }}}
  </body>
</html>
```
When running prettier via command line or in the [prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAeAhAEQPIGEAqAmgAoCiABABYwC2ANgHwA6Uq19zU55bcAhgBNO3bsDFV+A8gF9ZLEeTHByAJwQC4KgEpwAzhACuKsHAASAS1i6AFAEoZcriKWr1mgMowAnnT12H0vI8APSUksI8AEYQAl4RouJqUBoqnnzw-rKBTgnKSSnuYCrmAA4wNvZZQajB0bGcNeyMIAA0IBBl5tC6yKB8KioQAO7E-Qg9KHwAbhDmAq0gkSp8YADWcDDuJSuWAObIMCoGcG1NAOqU5vC62ybu41fmU1deyOC6PW2WupowxMu7Gh8ZAAMz4dB+bQAVroAB4AIWWaw27j4NDgABlLHBQeDISAYbD3HtfABFAwQeC4iEnEDbFQ-FRvJoLErFWBnOYwSjIAAcAAY2myID8zssSm82XpNFMcW01ABHAzmNT-PiA4FIME0to-Gjman43QkuDkyk4pCHY5tGB8SKcgTc5AAJhty3MdD2uAgNCBbz0AFYFgYfvg7RNtT9ZEA) you get the expected result .
```html
<!DOCTYPE html>
<html>
  <head>
    {{{ head }}} {{{ renderResourceHints() }}} {{{ renderStyles() }}}
  </head>
  <body>
    {{{ renderState() }}} {{{ renderScripts() }}}
  </body>
</html>
```

However when running pretter via VS Code the output is different
```html
<!DOCTYPE html>
<html>
  <head>
    {{{ head 
    }}} 
    {{{ renderResourceHints() 
    }}} 
    {{{ renderStyles() 
    }}}
  </head>
  <body>
    {{{ renderState() }}} {{{ renderScripts() }}}
  </body>
</html>
```
[...]

PR #810, opened 2019-04-24
Add a test case for html with literals

Archanium:
This is mainly adding a test case for https://github.com/prettier/prettier-vscode/issues/719


CiGit:
We don't resolve the parser the same way as prettier's CLI.
We don't check matching file extension.

Archanium:
I'd like to help with this if I could in any way.
Is there ways to see which parser was used for a specific file? (also for the prettier-cli?)

ntotten:
This is fixed in version 3.0. See: 7d16b917ee95a5267e97047a28f3678331166d72

You can install the preview here: https://marketplace.visualstudio.com/items?itemName=Prettier.prettier-vscode

PR #811, opened 2019-04-24
Resolve prettier to bundled if findPkg can't find it

### LINKED ISSUE #522: Use global `prettier` package as a fallback when Prettier is not installed inside a project.

visnup:
Fixes #522
Slightly different implementation of #561

### DIFF
--- a/src/requirePkg.ts
+++ b/src/requirePkg.ts
@@ -30,22 +30,23 @@ function findPkg(fspath: string, pkgName: string): string | undefined {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


CiGit:
Thank you!

PR #883, opened 2019-07-26
Use webpack to reduce the code size and loading times (8.5x gain of size on disk)

badre429:
This PR enables the usage of webpack to bundle the extension, so that the corresponding js produced and the code load times are lesser

**Old install directory:** 
> size:67,7 MB (71 011 275 bytes)  
> size on disk:78,9 MB (82 751 488 bytes)  
> contains :7 787 Files, 1 229 Folders  

**New install directory:**
> size:9,22 MB (9 672 392 bytes)  
> size on disk:9,23 MB (9 687 040 bytes)   
> contains :8 Files, 1 Folders


ntotten:
Thanks! Due to all the changes I replaced this with #902 902

Issue #870, opened 2019-07-09
[BUG] not working with eslint 6


jimblue:
Hi,

When using the option **eslint-integration**, Prettier is not working and output the following error:

```bash
ENOENT/home/dev/frontend/webpacker/app/index.js:: no such file or directory, open '/home/dev/projects/frontend/node_modules/eslint/lib/linter.js'
```

I've check in my **node_modules** and effectively, `linter.js` doesn't exist.

It was previously working with **eslint version 5** because `linter.js` did exist at this path [here](https://github.com/eslint/eslint/blob/5.x/lib/linter.js).

But in **eslint version 6**, **linter.js** has been moved in a subfolder [here](https://github.com/eslint/eslint/tree/master/lib/linter).

That's why it's actually broken.
So basically all you have to do is to modify the require:

**from:**
```js
require /node_modules/eslint/lib/linter.js
```

**to:**
```js
require /node_modules/eslint/lib/linter/index.js
```

Cheers

jakeleventhal:
Need this ASAP!

chengjia080787:
any luck on this? it's a critical issue to be honest

mbalaam:
I’ve ended up removing the extension and I am using https://github.com/prettier/eslint-plugin-prettier instead.

Thomazella:
running `"prettier.eslintIntegration": false` fixes

asbjornh:
> **to:**
> 
> ```js
> require /node_modules/eslint/lib/linter/index.js
> ```

Maybe change to `require('node_modules/eslint/lib/linter')` instead? Node's module resolution would then check both `linter.js` and `linter/index.js`. If the import looked like that in the first place we wouldn't have an issue now 😃

asbjornh:
@jimblue Where is that `require` call coming from? I can't see that this package imports `eslint` anywhere in the source and `eslint` isn't listed in `package.json` either. Actually I can't find the text `"eslint/lib/linter"` anywhere when grep'ing the plugin folder 😯

asbjornh:
I tried creating a fresh repo to reproduce, and I'm now getting a different error message:

```
Failed to load plugin 'prettier' declared in 'CLIOptions'<project-path>/index.js:: Cannot find module 'eslint-plugin-prettier'
Require stack:
- /__placeholder__.js
```

`eslint-plugin-prettier` is most definitely installed. For anyone interested in debugging this, here is a setup for reproducing this:

```
  "dependencies": {
[...]

PR #901, opened 2019-08-12
Removes Built-in Linters

ntotten:
This is a proposal for removing the linters in favor of using the official linter extensions along with their `autoFixOnSave` settings. 

- [x] Run tests


asbjornh:
This seems relevant :) 
https://github.com/microsoft/vscode-eslint/issues/600

Up until recently the linter integration has worked flawlessly and has really improved my everyday developer experience. My concern, given the above mentioned issue would be that the thing that I love would be replaced by an inferior thing that would annoy me.

Seeing as this change was at least partially motivated by https://github.com/prettier/prettier-vscode/issues/870 it seems worth noting that fixing that issue would probably just be a matter of waiting for a single dependency to be updated which isn't necessarily a cause for panic 😛

kylemh:
Hey Nathan! You asked us to let us know how you felt [here](https://github.com/prettier/prettier-vscode/issues/870#issuecomment-520557432).

I guess my only concern is how I should be handling any sort of race conditions between rules that may fight. Are you going to somehow guarantee your formatting runs before ESLint's or TSLint's?

PR #902, opened 2019-08-12
Bundle with webpack

ntotten:
This is a draft as there still are warnings on webpack to deal with that may cause issues.


ntotten:
Fixed in a9c1a86668fa4251e5a3a409df168d627c639915

PR #985, opened 2019-10-05
adds prettier-demo.gif

GoodGuyGregory:
- [ ] Run tests

**Readme Modifications**

- [x] Adds Demo of this awesome plugin for VS-code

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ ext install esbenp.prettier-vscode
 
 ## Usage
 
+![prettier-demo](./prettier-demo.gif)
+
 ### Using Command Palette (CMD/CTRL + Shift + P)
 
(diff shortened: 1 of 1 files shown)


kachkaev:
Not sure that bloating README by 6.2 MB is justified. Besides, the gif is not framed perfectly. Line numbers and the first line are cut, tab spacing it 8 characters (while it's 2 by default).

GoodGuyGregory:
I could recenter everything. I just feel like a demo would be easier for people to understand the immediate results of downloading and using the plugin. Sorry if it seemed redundant

ntotten:
Given that this will ship with every download of the extension, this seems like too much to put in.

PR #1065, opened 2019-11-16
Fix for running Prettier on files in node_modules dirs.

### LINKED ISSUE #1064: Error message whenever opening a file in `node_modules`.
### Summary

Since d19f93c5dae3e5ea0795a8490869d0bb69e629e0 in v3.0.0, whenever I open a file in `node_modules` that has its own `package.json` that has `prettier` in its deps, I get the error `Failed to load module. Attempted to load prettier from package.json`. This only happens if I don't have `prettier` installed as a deps in my own local project's `package.json`.

Error message comes from 470a9bfd74bcbbf1e41bdf8cbe11ee135bd0e8b9.

### Steps To Reproduce:

1. Install `prettier` as a global dep.
2. In `~/.prettierrc.json`, add some config, like:

```json
{
  "arrowParens": "always",
  "singleQuote": true,
  "trailingComma": "all"
}
```

3. Open a JS project, in which `prettier` is not a dep. 
4. Install and open a dep's `package.json` that has `prettier`. In my case, I had `@octokit/rest` installed as a dep in my project.
5. Open a file in the dep. So, open `node_modules/@octokit/rest/index.d.ts`, for instance.
6. The error appears: `Failed to load module. Attempted to load prettier from package.json`.

### Expected result

No error should appear. This extension should ignore `package.json` files in `node_modules` dirs, even if they contain `prettier` as a dep.

### Actual result

The error mentioned occurred.

### Additional information

**VS Code Version**: 1.39.2

**Prettier & Prettier Plugin Version**: Prettier 1.19.1, plugin 3.6.0

**OS and version**: macOS 10.14.6 (18G1012)

garyking:
- [x] Run tests

This is a fix for when opening a file in a `node_modules` dir, that also has `prettier` as a dep.

This closes #1064.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file. -->
 
-## 3.6.0]
+- Fix for finding local `prettier` module in packages located in `node_modules` dirs.
+
+## [3.6.0]
 
 - Added back status bar button
 
--- a/src/ModuleResolver.ts
+++ b/src/ModuleResolver.ts
@@ -172,8 +172,19 @@ export class ModuleResolver implements Disposable {
    * @returns {string} resolved path to module
    */
   private findPkg(fspath: string, pkgName: string): string | undefined {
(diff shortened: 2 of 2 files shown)


ntotten:
Thanks!

PR #1168, opened 2019-12-25
feat: add option disableRequireLocalPkgErrorMessage

zyy7259:
- [x] Run tests


zyy7259:
Sometimes I just want to clone a repo and browse the code without installing all the deps. And I don't want to see the error message every time I open a file.

ntotten:
I’m not sure I support this being a setting. Maybe we can make the error less frequent. I’ll think about it a bit.

zyy7259:
I think that's ok, thanks~

PR #1215, opened 2020-01-22
add snippets to include prettier-ignore comments

libinvarghese:
- [x] Run tests


ntotten:
Thanks for the contribution, but I would prefer to keep the extension as simple as possible. I'd suggest releasing this as a standalone extension.

Issue #1266, opened 2020-03-11
Detect Prettier module based on node_modules presence rather than package.json dep


JHilker:
**Is your feature request related to a problem? Please describe.**
Yes, but bare with me this may sound a bit weird. For my companies internal stack we're planning on distributing a shared version of Prettier (and it's configuration) across the entire org of >700 FE repos. This is primarily an effort to reduce the overhead of managing it's version / config / local installation, allowing us to rollout updates without sending out automated PRs to every repo.

As part of this change, the dependency will no longer be declared as a `package.json` dependency for these projects. However while doing so we want to ensure editor support doesn't break, so our setup ensures the necessary directories / binaries are still included in the `node_modules` directory and available for use.

**Describe the solution you'd like**
Currently it appears that the extension detects and resolves to a local version of Prettier based on finding a parent `package.json` with a direct dep/devDep on it. 

https://github.com/prettier/prettier-vscode/blob/cd1b3aa1f40f9fb5467a2cc74c2b35d78637de03/src/ModuleResolver.ts#L298-L306

I would propose enhancing this detection to also see if a `prettier` binary exists at `node_modules/.bin/prettier` relative to any discovered `package.json` paths. This check could be in addition to the current check, or a replacement to it.

**Describe alternatives you've considered**
The most obvious alternative here would be to specify `prettierPath`, but I'm hoping this could be avoided since it would need to be done in everyone's user or workspace VSCode settings.

**Additional context**
This appears to be how the `vscode-eslint` currently operates in order to find ESLint (which does work for our use case), so there is prior art in the community for this approach.

https://github.com/microsoft/vscode-eslint/blob/da0d3ce88482a14a2572b24aef75e554da6b4e66/client/src/utils.ts#L18-L27

I definitely realize that our use case is a bit non traditional, but I think the suggestion here could be a fairly general improvement to the detection of the local version of prettier. Looking at the other issues here, I think this may solve for https://github.com/prettier/prettier-vscode/issues/950 as well.

I'm happy to provide a PR for this, but wanted to run it by the team to make sure it made sense / there wasn't any blockers to it before starting the work. Open to alternative suggestions as well!

ntotten:
Yes, I agree this is something we should do. I haven't had time to look into it but I would love a PR!

PinkaminaDianePie:
[...]

PR #1291, opened 2020-04-03
1266: Detect Prettier based on node_modules presence

JHilker:
Resolves https://github.com/prettier/prettier-vscode/issues/1266

- [x] Run tests

This PR updates the ModuleResolver logic to recursively check parent directories for either a `package.json` containing the dependency, and now additionally the dependencies existence in the top level of the `node_modules` directory. If either is found, `findPkg` returns the resolved path based on that path.

Additionally, I've updated the `requireLocalPkg` logic to provide a better error notification message in the case that `findPkg` is unable to `resolve.sync` the dependency from the given directory. Previously it would default to `package.json`, but now it'll show the path it attempted to resolve it from. This typically would occur if the dependency appeared in the `package.json`, but was not yet installed.

image

I've tested this locally for both cases, any thoughts on the approach?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file. -->
 
+## [next]
+
+- Support loading Prettier from `node_modules` even if it doesn't appear as a direct dependency in a `package.json`
+
 ## [4.0.0]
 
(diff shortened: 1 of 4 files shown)


JHilker:
@ntotten, looks like my build failed due the classic:

```
info There appears to be trouble with your network connection. Retrying...
```

during the yarn install on Windows/Linux, and it looks like auth is required to rerun it. Mind pressing the button when have a chance, or letting me know how I can give it a go?

Thanks in advance for the review!

ntotten:
Awesome! Would you mind creating a test for this as well? I will merge and publish asap.

ntotten:
Moved to #1294 1294

PR #1294, opened 2020-04-05
Node mod presense

ntotten:
Replaces #1266


src/test/suite/module.test.ts
```diff
@@ -0,0 +1,16 @@
...
+
+suite("Test module resolution", function () {
+  this.timeout(10000);
+
+  test("it formats without prettier in package.json", async () => {
+    if (platform() === "win32") {
+      return assert.ok(true, "Skipping test on windows.");
+    }
+    const { actual } = await format("module", "index.js");
+    const expected = await getText("module", "index.result.js");
```
JHilker, line 13: Out of curiosity, does something in the test suite prevent this from using the Prettier version built into the extension? I figured a test case like this would actually pass prior to the changes I made as well because of that default.

yarn.lock
```diff
@@ -1702,7 +1702,7 @@ find-up@3.0.0, find-up@^3.0.0:
 
 find-up@^4.0.0, find-up@^4.1.0:
   version "4.1.0"
-  resolved "https://npm.hubteam.com/npm-nexus/repository/npm-all/find-up/-/find-up-4.1.0.tgz#97afe7d6cdc0bc5928584b7c8d7b16e8a9aa5d19"
+  resolved "https://registry.yarnpkg.com/find-up/-/find-up-4.1.0.tgz#97afe7d6cdc0bc5928584b7c8d7b16e8a9aa5d19"
```
JHilker: 😬. Sorry!

src/test/suite/module.test.ts
ntotten: Yeah, good point. I didnt even think about this. I need to change the test to check the version.

PR #1305, opened 2020-04-07
ModuleResolver: Add some unit tests

JHilker:
Follow up to https://github.com/prettier/prettier-vscode/pull/1294#discussion_r404040095. I couldn't figured out on a good way to test this with the existing more integrationy testing setup, so instead went with a unit-test approach. Curious what you think about this approach @ntotten 

Also adds a `.prettier-vscode-root` file, largely for testing purposes

- [x] Run tests
   * N/A


src/test/suite/ModuleResolver.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  suite("getPrettierInstance", () => {
+    test("it returns the bundled version of Prettier if local isn't found", () => {
+      const fileName = path.join(
+        getWorkspaceFolderUri("no-dep").fsPath,
+        "index.js"
+      );
+      const prettierInstance = moduleResolver.getPrettierInstance(fileName, {
+        showNotifications: true,
```
JHilker, line 33: Needing this preference seems a bit odd to me. It looks like it typically only relates to notifications, but in this case it prevents the log from showing. Is this intentional?

https://github.com/prettier/prettier-vscode/blob/ca5ea1d8dc2bb403db3be252c312801f8b7efbb2/src/ModuleResolver.ts#L117-L119

src/ModuleResolver.ts
```diff
@@ -334,6 +334,10 @@ export class ModuleResolver implements Disposable {
         if (fs.existsSync(path.join(dir, "node_modules", pkgName))) {
           return dir;
         }
+
+        if (fs.existsSync(path.join(dir, ".prettier-vscode-root"))) {
+          return findUp.stop;
+        }
```
JHilker, line 340: I'm not sure this would ever really be used outside of testing, so I didn't put it in the CHANGELOG. This just prevents the ModuleResolver logic from continuing to scan any parent directories if it's found, otherwise the tests would always fallback to the `prettier` definition in the root `package.json`, rather than the explicit defaulting logic.

Right now it's exclusively used for the `no-dep` case where it's required, but you think it makes sense, we may want to see about putting it in each of the workspace roots (or just in the `test-fixtures` directory itself?) as a follow up PR, to help ensure each workspace is testing against the version we expect to be.

PR #1334, opened 2020-04-16
Make node_modules resolution a fallback to explicit package.json resolution

JHilker:
Resolves https://github.com/prettier/prettier-vscode/issues/1266#issuecomment-614557540

Should handle both use cases pretty well. I imagine if there is ever an explicit package.json dep, you're in a repo/environment where you wouldn't want to resolve into an implicit one instead.

And even if you did you could technically leverage the `.prettier-vscode-root` file to enforce it doesn't continue parsing upwards past your implicit dep case.

Biggest con here is that we may parse the entire file tree twice now, but I don't think that should be too big a perf hit? I can dig into making it a single pass as a follow up PR if anyone's concerned (want to get the fix in to unblock the folks whose workflow broke)

- [x] Run tests

/cc @aju @PinkaminaDianePie @ntotten


PinkaminaDianePie:
Should work for my case, thank you for very fast response and fix! :)

JHilker:
```
UNRESPONSIVE extension host, 'esbenp.prettier-vscode' took 98% of 3782.286ms, saved PROFILE here: '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/exthost-078dba.cpuprofile' [
  { id: 'esbenp.prettier-vscode', total: 3705010, percentage: 98 },
```

@ntotten, is this something you have seen before, or think this reflects a problem in this PR? https://dev.azure.com/prettier/prettier-vscode/_build/results?buildId=6096&view=logs&j=7bab896a-24f8-544f-51eb-43745367a332&t=0e7f8c81-72e1-5201-9547-1d4d9915b15f&l=168

aju:
@JHilker any ideas about this one ?:)

JHilker:
I think we need to wait for @ntotten to have a chance to look at this and get it merged. He may be the only maintainer right now?

src/ModuleResolver.ts
```diff
@@ -331,6 +332,20 @@ export class ModuleResolver implements Disposable {
           }
         }
 
+        if (fs.existsSync(path.join(dir, ".prettier-vscode-root"))) {
```
ntotten, line 335: I'm not so sure about this. What is the purpose? VS Code already defines what a root of the project is, why not just use that?

ntotten:
@JHilker Yeah, I am the only maintainer right now. Always looking for help if you want to be more involved. Let me know.

src/ModuleResolver.ts
PinkaminaDianePie: What if you open not a project, but a subdirectory? You will have different root, so prettier can be resolved to a different version, it might be tricky for devs to debug such setup.

JHilker: This was actually introduced in my last PR (just copy pasted for the new code path of this PR). Here's more context on it: https://github.com/prettier/prettier-vscode/pull/1305#discussion_r405109478

TL;DR is that it was originally necessary for the tests, but could _technically_ have newfound usefulness in the scenario where you want to use an implicit dep instead of an explicit further up in the tree. I also agree with @PinkaminaDianePie that I'd be hesitant to depend on the VS Code root since it could break peoples workflows, esp in mono-repo environments.

Honestly I still don't think the flag would/should get usage outside of this project, which is why I elected to avoid documenting it. It's definitely a bit of a weird escape-hatch, so open to alternatives!

JHilker:
> Always looking for help if you want to be more involved. Let me know.

I'm definitely interested, and happy to support this project however I can! Feel free to let me know if there are ways I can help out.

src/ModuleResolver.ts
ntotten: I updated the code to strip this out in prod builds as it is for testing only: https://github.com/prettier/prettier-vscode/commit/1fe54f1134058cb684329dfa69ccda1716a07172

JHilker: Looks like that commit only handled one of two locations after this PR was also merged. Tossed up a PR to fix this and the failing mater tests: https://github.com/prettier/prettier-vscode/pull/1356

PR #1335, opened 2020-04-17
Add support for arm-template language, which is a superset of JSON

bergmeister:
Currently formatting an ARM template does not work any more since the ARM extension defined their own language: https://github.com/microsoft/vscode-azurearmtools/issues/465

I haven't tested this change, this is to start discussion if it would be as simple as that to re-add support for this language, which is just a superset of `jsonc`.

- [ ] Run tests

### DIFF
--- a/src/LanguageResolver.ts
+++ b/src/LanguageResolver.ts
@@ -56,7 +56,7 @@ export class LanguageResolver {
       "javascriptreact",
       "typescript",
       "typescriptreact",
-      "json",      
+      "json",
       "graphql",
     ];
   }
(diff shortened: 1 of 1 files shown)


ntotten:
This seems fine to me, please update the release notes.

bergmeister:
@ntotten Can you confirm that this change would make it work again? Unfortunately I could not find instructions how to build it locally and just pressing F5 in VS-Code did not work (I've installed node and did an `npm install`). It's also not helpful that your build does not upload the built extension as an artifact so that I could download it from the build and install locally for testing.

ntotten:
This change actually isn't going to fix the issue. You need to update Prettier itself by specifying `vsclodelanguageIds` for the custom language. https://github.com/prettier/prettier/blob/master/src/language-js/index.js#L39-L66

bergmeister:
@ntotten Thanks, I've tried to open a new PR, can you have a look please: https://github.com/prettier/prettier/pull/8274

PR #1428, opened 2020-06-12
Remove hardcoded prettier defaults for compatibility with earlier versions

### LINKED ISSUE #1290: prettier-vscode@4.0.0 uses prettier@2.x config by default
### Summary

prettier-vscode@4.0.0 uses prettier@2.x config by default

### Github Repository to Reproduce Issue

https://github.com/trivikr/prettier-vscode-4-test

### Steps To Reproduce:

- Clone this repo, and open it in VSCode
- Install [prettier-vscode extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) v4.0.0
- Run `yarn` to install dependencies
- Open [index.js](./index.js), and remove trailingComma from the end of array
- Notice that prettier-vscode extension adds the trailingComma on Save

### Expected result

prettier-config should use default config for the prettier version installed in the workspace

### Actual result

prettier-vscode@4.0.0 uses prettier 2 config.

<details>
<summary>Screen recording</summary>

prettier-vscode-adds-trailingComma

</details>

### Additional information

Issue in which prettier 2 config was not used with prettier-vscode@3.20.0: https://github.com/prettier/prettier-vscode/issues/1285

**VS Code Version**:
Version: 1.43.2
Commit: 0ba0ca52957102ca3527cf479571617f0de6ed50
Date: 2020-03-24T07:34:57.037Z
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Darwin x64 18.7.0

**Prettier Extension Version**:
Extension Name: "prettier-vscode".
Extension Version: "4.0.0".

**OS and version**:
macOS 10.14.6 (18G4032)

## Prettier Log Output

<details>
<summary>Details</summary>

```
["INFO" - 8:38:54 AM] Extension Name: "prettier-vscode".
["INFO" - 8:38:54 AM] Extension Version: "4.0.0".
["INFO" - 8:38:54 AM] Loaded module 'prettier@1.19.1' from '/Users/trivikr/workspace/prettier-vscode-test/node_modules/prettier/index.js'
["INFO" - 8:38:54 AM] Enabling prettier for languages
[
  "css",
  "graphql",
  "html",
  "javascript",
  "javascriptreact",
  "json",
  "json5",
  "jsonc",
  "less",
  "markdown",
  "mdx",
  "mongo",
  "postcss",
  "scss",
  "typescript",
  "typescriptreact",
  "vue",
  "yaml"
]
["INFO" - 8:38:54 AM] Enabling prettier for range supported languages
[
  "graphql",
  "javascript",
  "javascriptreact",
  "json",
  "typescript",
  "typescriptreact"
]
["INFO" - 8:38:57 AM] Formatting /Users/trivikr/workspace/prettier-vscode-test/index.js
["INFO" - 8:38:57 AM] Using ignore file (if present) at /Users/trivikr/workspace/prettier-vscode-test/.prettierignore
["INFO" - 8:38:57 AM] Loaded module 'prettier@1.19.1' from '/Users/trivikr/workspace/prettier-vscode-test/node_modules/prettier/index.js'
["INFO" - 8:38:57 AM] File Info:
{
  "ignored": false,
  "inferredParser": "babel"
}
["INFO" - 8:38:57 AM] No local configuration (i.e. .prettierrc or .editorconfig) detected, falling back to VS Code configuration
["INFO" - 8:38:57 AM] Prettier Options:
{
  "arrowParens": "always",
  "bracketSpacing": true,
  "endOfLine": "lf",
  "htmlWhitespaceSensitivity": "css",
  "insertPragma": false,
  "jsxBracketSameLine": false,
  "jsxSingleQuote": false,
  "printWidth": 80,
  "proseWrap": "preserve",
  "quoteProps": "as-needed",
  "requirePragma": false,
  "semi": true,
  "singleQuote": false,
  "tabWidth": 2,
  "trailingComma": "es5",
  "useTabs": false,
  "vueIndentScriptAndStyle": false,
  "filepath": "/Users/trivikr/workspace/prettier-vscode-test/index.js",
  "parser": "babel"
}
["INFO" - 8:38:57 AM] Formatting completed in 31.578346ms.

```

</details>

davidtaylorhq:
Some of Prettier's defaults changed in version 2.0, so the defaults in this vscode plugin were updated to match in c2e9c55f. This caused configuration to break for prettier 1.x projects without explicit config files. This commit adds a new 'default' option to each of these settings. With this new default, the plugin does not pass any value for that setting to prettier, allowing the version-specific default to be used.

If overrides are selected in the config, they are passed along to prettier as before.

This resolves #1290 using a different approach to 5baeeb6, and I think it should be more future-proof against prettier changing defaults again in future.

- [x] Run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file. -->
 
+## [Unreleased]
+
+- By default, allow prettier to control defaults itself. Overrides configured in the plugin settings will only be passed if non-default.
+
 ## [5.0.0]
 
 - Removed support for legacy linter integration. [See documentation](https://github.com/prettier/prettier-vscode#linter-integration) on how to configure linters.
--- a/README.md
+++ b/README.md
@@ -208,9 +208,7 @@ Disable format on save so this extension doesn't run and enable code actions to 
 
 ### Prettier Settings
 
-All prettier options can be configured directly in this extension. These settings are used as a fallback when no configuration file is present in your project, see the [configuration](#configuration) section of this document for more details. For reference on the options see the [prettier documentation](https://prettier.io/docs/en/options.html).
-
-> The default values of these configurations are always to their Prettier 2.0 defaults. In order to use defaults from earlier versions of prettier you must set them manually using your VS Code settings or local project configurations.
+All prettier options can be configured directly in this extension. These settings are used as a fallback when no configuration file is present in your project, see the [configuration](#configuration) section of this document for more details. If left unchanged, the prettier default will be used. For reference on the options see the [prettier documentation](https://prettier.io/docs/en/options.html).
 
 ```
 prettier.arrowParens
--- a/package.json
+++ b/package.json
@@ -187,134 +187,218 @@
           "scope": "resource"
         },
         "prettier.printWidth": {
-          "type": "integer",
-          "default": 80,
+          "type": [
+            "integer",
+            "null"
+          ],
+          "default": null,
           "markdownDescription": "%ext.config.printWidth%",
           "scope": "resource"
         },
         "prettier.tabWidth": {
-          "type": "integer",
-          "default": 2,
+          "type": [
+            "integer",
+            "null"
+          ],
+          "default": null,
           "markdownDescription": "%ext.config.tabWidth%",
           "scope": "resource"
         },
         "prettier.singleQuote": {
-          "type": "boolean",
-          "default": false,
+          "type": [
+            "string",
+            "boolean"
+          ],
+          "enum": [
+            "default",
+            false,
+            true
+          ],
+          "default": "default",
           "markdownDescription": "%ext.config.singleQuote%",
           "scope": "resource"
         },
         "prettier.trailingComma": {
           "type": "string",
           "enum": [
+            "default",
             "none",
             "es5",
             "all"
           ],
-          "default": "es5",
+          "default": "default",
           "markdownDescription": "%ext.config.trailingComma%",
           "scope": "resource"
         },
         "prettier.bracketSpacing": {
-          "type": "boolean",
-          "default": true,
+          "type": [
+            "boolean",
+            "string"
+          ],
+          "enum": [
+            "default",
+            true,
+            false
+          ],
+          "default": "default",
           "markdownDescription": "%ext.config.bracketSpacing%",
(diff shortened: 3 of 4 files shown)


ntotten:
My stance is that every project should have a prettier config file. Defaults are there for simplicity on starting new projects. Any project that “breaks” because of an extension version changes is improperly configured.

davidtaylorhq:
> Defaults are there for simplicity on starting new projects.

That makes sense, although I worry that they have actually become a **complication** on starting new projects. The defaults shipped with this plugin do not necessarily match those of the installed prettier version. At the moment adding an **empty** config file causes a change in behaviour, which seems like an unusual UX to me.

In our case at Discourse, we have hundreds of official and community themes/plugins. We have been seeing commits from people using this VSCode plugin which violate prettier rules when checked by our CI pipelines. This has created a lot of extra work to remedy the formatting and teach users how to update their VSCode config. Granted, we could add empty prettier config files to all those repositories as a workaround, but the fundamental issue is that this plugin does not work the same as running prettier on the CLI.

This PR would make it work consistently, regardless of prettier version, and without losing any functionality. Alternatively...

> My stance is that every project should have a prettier config file

How would you feel about completely removing the default choices from the VSCode preferences? If people want to change the config, they should use a config file.

Doing that would resolve the inconsistency, be future-proof against future prettier default changes, and also encourage people to follow best practices when using prettier.

ntotten:
Maybe another alternative is to NEVER use VS Code configuration if a local version of prettier is installed, that way we always use the defaults from prettier - either through config or just accepting the defaults of the local version. But I am not sure this would help you in this case - do all of these themes have local prettier?

I get the pain this can cause and am certainly open to discussions but this has been brought up many times and I have yet to find anything that resolves all of the different combinations. The simple thing from my perspective is to just add the config.

Maybe we can start warning people when we fallback to vscode config and encourage them to add a prettier config file.

I once tried to remove the VS Code config entirely because people actually can set global defaults on their machine by adding a `.prettierrc` to their home directory but that went over poorly with people. With 1m MAU of this extension changes like this can be extremely difficult. :)

frol:
I agree with @davidtaylorhq. This extension is playing against the core values of prettier of being a zero-config workflow ensuring consistent formatting in any environment. Falling back to something that prettier does not respect itself is a fail. I see lots of devs bumping into this issue arguing that they use "prettier" while they in fact are using "prettier-vscode" instead of "prettier". It is causing completely unnecessary conversations and workarounds.

@ntotten Please, reconsider your decision.

PR #1905, opened 2021-04-15
Ignore `requirePragma` in forced mode

dtcristo:
A feature request with implementation all in one.

## Feature request

**Is your feature request related to a problem? Please describe.**
If my project has `requirePragma: true` set in prettier config, there is currently no way to "force" formatting of the file without first manually adding the pragma comment.

**Describe the solution you'd like**
I would like the recently added "Format Document (Forced)" command to ignore the `requirePragma` setting and format the file anyway. It is called **Forced** after all, shouldn't this action take precedence?

**Describe alternatives you've considered**
I've not really considered other approaches, but open to suggestions on better ways to solve my problem.

## Implementation

We ensure `requirePragma` is `false` if the execution is running in forced mode and `requirePragma` is set `true` in config.

As you can see, I also did a small refactor of `rangeFormattingOptions` as it made the signature for `getPrettierOptions` cleaner.

- [x] Run tests – `it uses config from vscode settings` **fails** for me, but that test also fails on `main`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](http://keepachangelog.com/) for recommendations on how to structure this file. -->
 
+## [Unreleased]
+
+- Forced mode now ignores `requirePragma` config
+
 ## [6.3.2]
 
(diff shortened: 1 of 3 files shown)


azz (approved):
:+1: lgtm

ntotten (approved):

PR #1946, opened 2021-05-10
Set ^1.56.0 as VS Code engine

aravindvnair99:
Fixes #1944

- [x] Run tests


ntotten:
This isnt a bug. There is no reason to increase the VS Code API version unless we need those features.

PR #2059, opened 2021-07-02
Update and reorganize .vscodeignore

tjquillan:
This updates .vscodeignore removing unused entries and adding some new ones. Additionally, reorganized to follow the order of a file explorer. A `/` was also added to the end of `node_modules` because while it is not required anymore as of `vsce@1.95.0` because they [reverted the switch to `ignore`](https://github.com/microsoft/vscode-vsce/issues/588) it seemed wise to protect against similar changes in the future and doesn't change the current functionality.

Added:
* .husky/
* .gitattributes
* .nvmrc
* .prettierrc

Removed:
* samples/
* azure-pipelines.yml

- [x] Run tests
- [x] Update the [`CHANGELOG.md`][1] with a summary of your changes (**I did not as this is not a public facing change**)


.vscodeignore
```diff
@@ -1,23 +1,24 @@
...
 .gitignore
-.github/
+.nvmrc
+.prettierignore
+.prettierrc
+CODE_OF_CONDUCT.md
+CONTRIBUTING.md
+CONTRIBUTORS.md
 tsconfig.json
+webpack.config.js
```
and-bulk, line 18: why? :)

.vscodeignore
```diff
@@ -1,23 +1,24 @@
...
 yarn.lock
-CONTRIBUTORS.md
-CONTRIBUTING.md
-CODE_OF_CONDUCT.md
 
-node_modules
+node_modules/
```
and-bulk, line 21: I think not needed `/` after dir

.vscodeignore
tjquillan: Can you clarify? If you mean why is the `webpack.config.js` there, then that would be because it shouldn't be included in the final bundle (and it was already there before this PR).

.vscodeignore
tjquillan: It is not needed but other directories had it so I added it for consistency and additionally it makes it clear it is a directory and not a file.

PR #2180, opened 2021-09-13
add web support

aeschli:
Hi prettier-vscode team!

:wave: I'm Martin and I'm a dev in the VS Code team. We recently launched VS Code for the web with [github.dev](https://docs.github.com/en/codespaces/developing-in-codespaces/web-based-editor).

In VS Code for the web, both UI and extension host run inside the browser. 

prettier-vscode was one of our test cases, and this PR shows how we got it to run. I hope you are interested to adopt and the PR is useful. I'm not an expert on prettier, be warned. I don't think you want the merge the PR as is, it's more for inspiration.

Web extension have a different main file than regular extensions. It's listed under the `browser` property. Web extensions can't load any modules except `vscode`. So all needs to be packaged and statically bundled.

I added new section to the webpack file for the web main entry file. It uses the same main source file but replaces the regular ModuleResolver with a `BrowserModuleResolver` which uses `prettier/standalone`. 

I added some abstractions for the ModuleResolver and the prettier types. Check out the types.d.ts file.

The PrettierEditService can then work on these abstractions. That was the idea. It's all a bit complicated, so maybe the idea of reusing PrettierEditService isn't the best one.

Note that in the web, the file system is virtual (not-file), so I removed all the `file` and `untitled` requirements.
So as a benefit these changes can also be used to run prettier-vscode on virtual workspaces.

I'm happy to give more information if needed.

To try it out:
- `yarn`
- run the new launch config `Launch Web Extension`. This will open a regular VS Code Desktop but with the prettier-vscode extension running in a web worker. To verify this, use the extension and run command `Show Running Extensions`. It says `web worker` next to `prettier.vscode`
- try it out in a browser with: `yarn chrome`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ out
 dist
 node_modules
 .vscode-test
+.vscode-test-web
 *.vsix
 .DS_STORE
 *.log
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -39,6 +39,18 @@
       ],
       "outFiles": ["${workspaceFolder}/out/test/**/*.js"],
       "preLaunchTask": "npm: test-compile"
+    },
+    {
+      "name": "Run Web Extension",
+      "type": "pwa-extensionHost",
+      "debugWebWorkerHost": true,
+      "request": "launch",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceFolder}",
+        "--extensionDevelopmentKind=web"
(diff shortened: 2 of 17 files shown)


tanhakabir:
cc @ntotten

ntotten:
Thanks! I'll take a look and merge asap.

ntotten:
Moving to #2203

PR #2312, opened 2021-12-03
Adding to read me

CaroEvans:
I had this extension for a while and didn't know that you could auto format on save until someone told me.
I think it would help beginners to have it in the intro paragraph.


ntotten:
This isn't really a feature of the extension though as it's just a VS Code setting that applies to any extension.

CaroEvans:
Doesn't that apply to a lot of stuff in this readme? @ntotten

PR #2438, opened 2022-02-21
Generate the config file using the extension settings

ionut-botizan:
- [x] Run tests

---

Currently, the `"Create Configuration File"` command generates a config file with only 2 hard-coded rules (`tabWidth` & `useTabs`).

This PR changes the initial content the command uses to generate the config file so that it uses the formatting rules defined in the VSCode settings. This way, if a user has customized their default Prettier settings in VSCode, those will also be used as a starting point when generating a `.prettierrc` file in a new project.

Alternatively, if you prefer not to change the behavior of the existing command, an additional command could be added that explicitly generates the file using the VSCode settings (A bit too verbose IMO, but maybe something like `"Create Configuration File Using the Extension's Settings"`?).

What do you think?

### DIFF
--- a/src/PrettierEditService.ts
+++ b/src/PrettierEditService.ts
@@ -24,7 +24,11 @@ import {
   PrettierOptions,
   RangeFormattingOptions,
 } from "./types";
-import { getConfig, getWorkspaceRelativePath } from "./util";
+import {
+  getConfig,
+  getWorkspaceRelativePath,
+  filterFormattingOptions,
+} from "./util";
 
 interface ISelectors {
   rangeLanguageSelector: ReadonlyArray<DocumentFilter>;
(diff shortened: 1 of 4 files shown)


src/PrettierEditService.ts
```diff
@@ -454,23 +458,7 @@ export default class PrettierEditService implements Disposable {
...
-      vsOpts.proseWrap = vsCodeConfig.proseWrap;
-      vsOpts.quoteProps = vsCodeConfig.quoteProps;
-      vsOpts.requirePragma = vsCodeConfig.requirePragma;
-      vsOpts.semi = vsCodeConfig.semi;
-      vsOpts.singleQuote = vsCodeConfig.singleQuote;
-      vsOpts.tabWidth = vsCodeConfig.tabWidth;
-      vsOpts.trailingComma = vsCodeConfig.trailingComma;
-      vsOpts.useTabs = vsCodeConfig.useTabs;
-      vsOpts.vueIndentScriptAndStyle = vsCodeConfig.vueIndentScriptAndStyle;
+      Object.assign(vsOpts, filterFormattingOptions(vsCodeConfig));
```
ntotten, line 461: Im pretty sure there is a reason we set these manually. I think there are extra values in some cases although I'm on my phone now and can't recall. Can you double check?

PR #2448, opened 2022-03-02
Allow formatting ranges in jsonc

rotu:
- [X] Run tests
- [X] Update the [`CHANGELOG.md`][1] with a summary of your changes


rotu:
I can't figure out why `allRangeLanguages` even exists. Why don't all supported languages allow range formatting? Is it still necessary?

rotu:
@ntotten, you introduced this whitelist. Could you weigh in?

ntotten:
I didn't actually introduce it, I probably moved the code last. Prettier used to only support range formatting on certain languages, if that isn't true anymore happy to take a PR (with tests) removing this flag.

rotu:
Ahhh. Gotcha. I tested it, and it seems that, if prettier doesn't support range formatting in a file, it will just ignore the `--range-start` and `--range-end` options and format the whole file.

Issue #1277, opened 2020-03-25
Run eslint --fix AFTER prettier on save


protoEvangelion:
**Describe the solution you'd like**
Due to company constraints, I'd like to run `eslint --fix` AFTER this extension runs Prettier.

**_Now that the prettier eslint integration is deprecated, is there a recommended way to do this_**? It looks like this has been brought up in the past: https://github.com/prettier/prettier-vscode/issues/14

**Describe alternatives you've considered**
I'm using this configuration, however it runs eslint FIRST & then Prettier:

```
"editor.codeActionsOnSave": {
    "source.fixAll": true,
    "source.fixAll.eslint": true
  },
```

ntotten:
Unfortunately, this is not possible. The order of operations is controlled by VS Code and formatting is not exposed as a code action. There are multiple issues in vscode about this and hopefully they will resolve soon. One example is here: https://github.com/microsoft/vscode/issues/90220

rohit-gohri:
The march VS Code update allows specifying an array to order codeActionsOnSave: https://code.visualstudio.com/updates/v1_44#_explicit-ordering-for-code-actions-on-save

rohit-gohri:
> The march VS Code update allows specifying an array to order codeActionsOnSave: https://code.visualstudio.com/updates/v1_44#_explicit-ordering-for-code-actions-on-save

So this and https://github.com/prettier/prettier-vscode/issues/870#issuecomment-571163060 gave me an idea for an extension, what if we could run 'Format Document' as a `codeActionOnSave`?

Checkout https://github.com/rohit-gohri/vscode-format-code-action/

This registers a custom codeAction: `source.fixAll.format`, which we can use instead of `formatOnSave`. It'll use whatever default formatter you have setup.

I'm running eslint after prettier with these settings:

```
  "editor.formatOnSave": false,
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.codeActionsOnSave": [
    "source.fixAll.format",
    "source.fixAll.eslint"
  ]
```

and these extensions:
```
[
    "esbenp.prettier-vscode",
    "dbaeumer.vscode-eslint",
    "rohit-gohri.format-code-action"
]
```

I hope this helps anyone coming here.

PR #2500, opened 2022-04-03
Add `prettier.forceFormatDocument` to code actions

### LINKED ISSUE #1555: Add code action to format with prettier in codeActionsOnSave in VS Code
Originally posted by @rohit-gohri in https://github.com/prettier/prettier-vscode/issues/1277#issuecomment-621175180 is a working solution for running prettier before eslint.

The above solution runs the default formatter (which can be set to Prettier) then uses the code action for eslint afterwards.

[There is already one example of a code action to create a config file in this extension](https://github.com/prettier/prettier-vscode/blob/main/src/commands.ts).

Could a code action be added for formatting? Such that if a user were to hit ctrl+shift+p in VS Code it would show something like "Prettier: Format Document"

[Here are examples from Microsoft](https://github.com/microsoft/vscode-extension-samples/tree/master/code-actions-sample)

roottool:
- [x] Run tests

Resolves #1555

## Overview

This PR adds two changes

- Add `prettier.forceFormatDocument` to code actions
- Adapt `settings.json` to this code action addition

### Adapt `settings.json` to this code action addition

I reset `editor.defaultFormatter` and turn off `editor.formatOnSave`. Because this extension is able to format by `source.prettier.forceFormatDocument` in `editor.codeActionsOnSave` without the two items.

## Run tests screenshot

Run tests screenshot


wedi:
@roottool Thank you for taking the initiative to get this long awaited feature in. 🎉 

Is there a special reason why you chose to name the action `source.prettier.forceFormatDocument`? I am asking because it looks like there is a de facto standard to use `fixAll.<productname>`. [This is used](https://github.com/microsoft/vscode-eslint/blob/fb60e3f2591db0d9e9d1bcfb45df5aeedb5111f4/client/src/settings.ts#L69) in Microsoft's own eslint extension and stylelint [mentions in its docs](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint&ssr=false#user-content-%60editor.codeactionsonsave%60) that you can enable all auto fixing by setting `"source.fixAll": true`.

roottool:
@wedi I forgot that I wrote the reason, sorry. It has two reason.

First, I wanted to be able to explicitly specify the order of execution. Because I don't know the execution order within `"source.fixAll": true`.
A `codeaction` that is added by this PR must be able to run from Prettier to AutoFix by ESLint in that order.
   > The above solution runs the default formatter (which can be set to Prettier) then uses the code action for eslint afterwards.
#1555

Second, I have decided that [`Source`](https://code.visualstudio.com/api/references/vscode-api#CodeActionKind) is more appropriate. Because I have determined that a statement ignoring the formatter's settings is not equivalent to a linter error.

> Fix all actions automatically fix errors that have a clear fix that do not require user input.
[VS Code API Docs - SourceFixAll](https://code.visualstudio.com/api/references/vscode-api#CodeActionKind)

And [this comment](https://github.com/prettier/prettier-vscode/issues/1555#issuecomment-1086888406) encouraged me to use `Source`.

> For exposing actions for the [CodeAction](https://code.visualstudio.com/api/references/vscode-api#CodeAction)sOnSave option, the plugin needs to provide a Code Action by using the CodeAction API, and especially a [CodeActionKind](https://code.visualstudio.com/api/references/vscode-api#CodeActionKind) of type "source"
https://github.com/prettier/prettier-vscode/issues/1555#issuecomment-1086888406

FYI: This `codeaction` names were appropriated from the existing command name. Since this `codeaction` uses that the existing command.
https://github.com/prettier/prettier-vscode/blob/991a53f27df9f6d2abc782d8065512310c120983/package.nls.json#L3
If you have a good name for it, I'd be glad to hear it!

wedi:
@roottool using the standard naming and choosing a custom execution order is not mutually exclusive. You can customize the order of execution by using the full names like this:

`settings.json`:
```yaml
  "editor.codeActionsOnSave": [
    "source.organizeImports",
    "source.fixAll.eslint",
    "source.fixAll.stylelint"
  ],
```

As prettier fixes all source [code] styling issues, `source.fixAll.prettier` sounds reasonable to me but I am not further involved in this project. I just wanted to point out that there is a naming convention which you did not follow.

Personally, I'll be happy with whatever name gets chosen and I'm exited to see this merged. :)

.vscode/settings.json
```diff
@@ -1,10 +1,10 @@
 // Place your settings in this file to overwrite default and user settings.
 {
   "editor.codeActionsOnSave": {
+    "source.prettier.forceFormatDocument": true,
```
ntotten, line 4: Is this a new recommended best practice from the VS Code team or just preference?

src/CodeActionOnSave.ts
```diff
@@ -0,0 +1,35 @@
...
+  CodeActionKind,
+  type CodeActionProvider,
+  type ProviderResult,
+} from "vscode";
+
+const FORCE_FORMAT_DOCUMENT_COMMAND = "prettier.forceFormatDocument";
+const FORCE_FORMAT_DOCUMENT_CODE_ACTION_KIND = CodeActionKind.Source.append(
+  FORCE_FORMAT_DOCUMENT_COMMAND
+);
+const FORCE_FORMAT_DOCUMENT_TITLE = "Format Document (Forced)";
```
ntotten, line 12: Why is it forced?

ntotten (changes requested):
A few comments

src/CodeActionOnSave.ts
roottool: Because this code action uses a command of the same name that exists in the command palette.
But I would be willing to adopt any better name. 👍

.vscode/settings.json
roottool: No, it's not. I'm going to undo this difference if it is not necessary.

This difference has two intent.
First, The intent of this difference is to resolve #1277 by realizing the following quoted portion.

> Originally posted by @rohit-gohri in https://github.com/prettier/prettier-vscode/issues/1277#issuecomment-621175180 is a working solution for running prettier before eslint.
> 
> The above solution runs the default formatter (which can be set to Prettier) then uses the code action for eslint afterwards.
> https://github.com/prettier/prettier-vscode/issues/1555#issue-702228085

Second, it is for testing the created new code action in this PR.
The intention was to make it clear that I was testing through the code action.

roottool:
I replied to a few comments. But I'm sorry that it took me so long to reply.

ntotten:
I kind of missed the "forced" part of this PR. I guess I am confused what the point of this is? Why would you want to always force prettier to format files that you have set to be ignored?

nujhong:
I think the contributor was re-using an existing command in the project as he mentioned https://github.com/prettier/prettier-vscode/blame/main/package.nls.json#L3

If I understood it correctly, the original commit https://github.com/prettier/prettier-vscode/commit/dcc531f6382fe0580edbf62386fc48ca233e9c2e was aimed to provide a way to "Format Document (Forced)" which means 👉 **format document specifically with formatter `esbenp.prettier-vscod`**

Therefore, the goal of this PR was to re-use the same command so that it can be registered as a **CodeAction** and allowing users to specify the order of code actions by preference.

This ultimately solves the good old "_Run Prettier First, then Run ESLint_" problem as an alternative approach to doing it via CLI commands or `eslint-plugin-prettier`.

Screen Shot 2022-07-25 at 9 45 24 pm

Looking forward to this being merged!

roottool:
@nujhong Thank you for the explanation assistance!

@ntotten This CodeAction name includes "forced" is because I uniformed the name at an existing command name. The CodeAction re-uses "Format Document (Forced)" command already in this extension.
https://github.com/prettier/prettier-vscode/blob/b395990a96e1cbd754646de0a2f21c4b36593cd3/package.nls.json#L3

Including the appropriation of this name, the implementation intent is as he commented.
I have no intention of enforcing the format on me. I just appropriated the name.

.vscode/settings.json
yamatsum: It looks like Microsoft's recommended settings

> Although you can also use the formatter on save using the setting editor.formatOnSave it is recommended to use the editor.codeActionsOnSave feature since it allows for better configurability.

https://github.com/microsoft/vscode-eslint

nd0ut:
Any updates?

ntotten:
I am going to close this. This PR doesn't work and it doesn't have tests. If somebody is interested in contributing this please continue the discussion here: https://github.com/prettier/prettier-vscode/issues/1555

PR #2604, opened 2022-06-28
Fix support for using a globally installed version of prettier

ehoogeveen-medweb:
- [x] Run tests

The extension's support for using a globally installed version of prettier wasn't working for me. I looked into it and found that command `npm.packageManager` was returning an empty value. It appears that you're actually supposed to pass the workspace folder URI as a parameter for this command to work!

While working on this pull request I also noticed that I was getting the error `command 'npm.packageManager' not found`. It appears this command comes from the `npm` extension. The extension is built into vscode, but for some reason it didn't get loaded. I found that adding this extension to the `extensionDependencies` section in `package.json` made the error go away.

Running the tests locally I get a few errors, but I get the same errors without my changes so I don't think they're related.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,6 +129,9 @@
     "semver": "^7.3.7",
     "vscode-nls": "^5.0.0"
   },
+  "extensionDependencies": [
+    "vscode.npm"
+  ],
   "capabilities": {
     "virtualWorkspaces": true,
     "untrustedWorkspaces": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -129,6 +129,9 @@
     "semver": "^7.3.7",
     "vscode-nls": "^5.0.0"
   },
+  "extensionDependencies": [
+    "vscode.npm"
```
ntotten, line 133: Will this cause any issues for people that are using yarn, etc?

ehoogeveen-medweb: As far as I know, package managers just ignore properties in package.json that they don't know about. [This](https://stackoverflow.com/a/27232456/3507557) StackOverflow answer also seems to indicate that this is generally valid. `extensionDependencies` is part of VSCode's extension manifest specification: https://code.visualstudio.com/api/references/extension-manifest

Issue #2402, opened 2022-01-30
requireConfig only if there's prettier config, and then also use .editorconfig


orenklein:
**Description of feature**
requireConfig of only prettier config, and if it exists then also use .editorconfig (If editorconfig exist but prettier config does not exist, skip the file) 

**Is your feature request related to a problem? Please describe.**
I have a monorepo with multiple project and a root editorconfig and I've also defined `prettier.requireConfig: true`. In a project that has prettier, formatting with prettier works perfectly and merges the root's editorconfig with the project prettierrc. However, few of the projects does not have a prettier config file, but they still get formatted with prettier (and its default settings) due to the root's editorconfig. 

If configuring `prettier.useEditorConfig: false`, it solves the issue for the projects without prettier config, but the projects with prettier config does not use the editorconfig for formatting anymore. 

**Describe the solution you'd like**
Perhaps adding another option to `requireConfig` named `prettierOnly`

**Describe alternatives you've considered**
Moving .editorconfig up to the specific projects and remove the root's editorconfig. You'd need to copy it to any new project, and changes will need to be copied across all the projects

ntotten:
I like the idea. I think we'd need to add a new config `requireLocalConfig` with values `never`, `prettierOnly`, `prettierAndEditorConfig` or something.

Happy to take a PR.

papandreou:
This is a much better default in my opinion. I originally specified the `requireConfig` option to _prevent_ this extension from kicking in when I'm working on eg. an open source project that might not have prettier enabled. There are way more projects that have an `.editorconfig` than a prettier configuration, so it's really annoying to suddenly get everything reformatted with prettier when I work on one of those.

papandreou:
From playing around with "Install another version" in vscode, this appears to be a regression introduced in 9.0.0. When I go back to 8.2.0, it works fine and just logs:

```
["INFO" - 9:16:32 AM] Extension Name: esbenp.prettier-vscode.
["INFO" - 9:16:32 AM] Extension Version: 8.2.0.
["INFO" - 9:16:34 AM] Formatting /home/andreas/work/express-validation/lib/index.js
["INFO" - 9:16:34 AM] Require config set to true and no config present. Skipping file.
```

whereas with 9.0.0 it goes:

```
["INFO" - 9:18:36 AM] Extension Name: esbenp.prettier-vscode.
["INFO" - 9:18:36 AM] Extension Version: 9.0.0.
[...]

PR #2708, opened 2022-09-08
Require a prettier config when requireConfig is true

redoPop:
Fix for #2402. When `requireConfig` is true we cannot rely on the `resolvedConfig` value to indicate that a Prettier config was found: the resolved configuration may have been picked up from an `.editorconfig` file instead.

This issue was [previously fixed in 8.1.0](https://github.com/prettier/prettier-vscode/blob/f4069b312e09fcdec9a77e882de2a0fe719f7f84/CHANGELOG.md#810) but the fix seems to have been undone by subsequent changes.

- [x] Run tests


ntotten:
I don't think this is right. I need to test, but I think `configPath` is just the location but doesn't necessarily indicate that there is a config at that location.

redoPop:
`configPath` is the resolution of [`prettier.resolveConfigFile`](https://prettier.io/docs/en/api.html#prettierresolveconfigfilefilepath) given the current document path. That's `null` if no Prettier config file is found in the parent tree, [coalesced to undefined in our code](https://github.com/prettier/prettier-vscode/blob/3af5f2c432c6b1be623fb67083e6c758c0689864/src/ModuleResolver.ts#L270).

One thing to watch out for as you test: if you have a prettier config in your home dir or elsewhere in the parent tree then it'll get picked up by resolveConfigFile. That seems like the right behavior for this feature, but it's a potential red herring for testing.

ntotten:
I'll merge this for v10 with prettier 3 as its a breaking change

redoPop:
> This pull request has been labeled as stale due to inactivity. Reply to keep this pull request open.

Keep open for next major release.

ntotten:
Actually, now that i think about this. Why can't you just set the config `prettier.useEditorConfig` to false? This will prevent the use of editorconfig as the config file and thus will require only a prettier config to be present.

redoPop:
@orenklein gave a good example in #2402: they're working in a monorepo with a global .editorconfig. Some of the individual packages in the monorepo augment those rules with Prettier configs, while others rely entirely on EditorConfig.

My team has a similar use case: we have a boilerplate .editorconfig in every repo to set broad formatting rules, and .prettierrc as an additional layer of configuration for projects that use Prettier.

In both use cases, it's helpful that Prettier allows us to extend EditorConfig rules and establish a hierarchy with a single source of truth for configurations that apply across multiple globs.

A couple of other factors I can think of:

* I work with projects from multiple teams, and it's useful to know that prettier-vscode will follow another team's Prettier config as best it can – including `editorconfig: true` – without requiring me to make additional configurations specific to that project.
* It's caught a few of us by surprise that EditorConfig satisfies the `requireConfig` rule. That doesn't necessarily make it a bug, but I think the natural expectation is that prettier-vscode requires a Prettier config, not necessarily an EditorConfig.

ntotten:
I think you’re points are valid, but from the history of this debate I think some people are going to want the current behavior. I don’t love extra complexity, but I wonder is we should evolve this config into maybe three values. 

```
useEditorConfig: “never” | “fallback” | “always”
```

the next challenge will be picking the names of these config in a way that makes sense. Open to suggestions.

redoPop:
A suggestion for config naming: we could change the values for _requireConfig_ instead of useEditorConfig. Something like this:

`"requireConfig": "strict" | "inclusive" | "none"`

* `strict` = Files are formatted only if a Prettier configuration file is present
* `inclusive` = Files are formatted if Prettier or EditorConfig configuration files are present
* `none` = Formatting is always enabled (default)

We could then leave useEditorConfig as is and let people opt in to the new behavior explicitly while continuing to support people who've come to expect the current behavior.

redoPop:
Keep open.

redoPop:
Keep open.

redoPop:
Keep open.

redoPop:
@ntotten What's the next step here? Happy to try implementing either of the proposed config changes in this PR, but not sure which you'd prefer.

redoPop:
Keep open?

redoPop:
Keep open?

PR #2852, opened 2023-01-28
Do not ship `.devcontainer` folder

mrazauskas:
Currently `.devcontainer` folder is not included in `.vscodeignore`, hence it is shipped with the extension.


mrazauskas:
Never mind.

PR #3027, opened 2023-06-14
fix: plugins `parser` not inferred

### LINKED ISSUE #3026: `prettier.getFileInfo` should pass `plugins` to help infer the plugin's `inferredParser`.
**Is your feature request related to a problem? Please describe.**

There is no `plugins` parameter passed now, so third-party plugins cannot correctly infer the `parser`, and formatting will be ignored. You must configure the correct `parser` in the configuration file's `overrides`, but this is cumbersome.
https://github.com/u3u/prettier-config/blob/e2ca8e5bebf2508230712497c61af3a5ee5b1fc4/src/index.ts#L21-L61

https://github.com/prettier/prettier-vscode/blob/d5ac67aa66822641491039a53dbe6388e5cd2594/src/PrettierEditService.ts#L420-L424

Before passing `plugins`:

image

After passing `plugins`:

image

**Describe the solution you'd like**

I think the plugin list should be automatically obtained and passed to the `plugins` parameter, like this:
https://github.com/u3u/prettier-config/blob/main/test/format.ts

```ts
import fs from 'node:fs/promises'
import path from 'node:path'
import prettier from 'prettier'
import { require } from '../src/utils/require'

export const format = async (file: string, config = '@u3u/prettier-config') => {
  const filepath = path.resolve('test/__fixtures__', file)
  const options = await prettier.resolveConfig('', { config: require.resolve(config) })

  const info = await prettier.getFileInfo(filepath, {
    // https://github.com/prettier/prettier-vscode/blob/d5ac67aa66822641491039a53dbe6388e5cd2594/src/PrettierEditService.ts#L420
    plugins: options?.plugins as string[],
    resolveConfig: true,
  })

  console.log(file, info)

  const buffer = await fs.readFile(filepath)
  const code = buffer.toString('utf8')

  const result = prettier.format(code, {
    ...options,
    filepath,
    parser: info.inferredParser || undefined,
  })

  return result
}
```

**Describe alternatives you've considered**

Users can only manually configure the correct `parser` for `overrides`, which is quite troublesome because you need to list the suffixes supported by the plugin.

https://github.com/u3u/prettier-config/blob/e2ca8e5bebf2508230712497c61af3a5ee5b1fc4/src/index.ts#L21-L61

Then I also found that if `prettier.format` does not pass `parser`, but only passes `filepath` and the `options` returned by `prettier.resolveConfig`, most parsers can correctly infer. I tested that only `.blade.php` will be inferred as `.php`.

**Additional context**

https://github.com/u3u/prettier-config

u3u:
## Fixed #3026

- Add test
- ~~Upgrade PNPM version to `v8`~~

~~After running `yarn test` without changing anything locally, I had 2 test failures. But I don't know why. Is it a problem with my local environment?~~

- [x] Run tests


u3u:
image

After confirmation, it was found that the reason for the failure of the following two tests was due to the existence of `~/.prettierrc.cjs` file in my local environment. After temporarily renaming it, all tests have passed.

- it uses config from `.editorconfig` file
- it uses config from vscode settings

In addition, because I use `8.6.2` for my local PNPM version, the version of lock file `test-fixtures/plugins-pnpm/pnpm-lock.yaml` will be updated. However, PNPM 8 requires Node.js 16 or above, while GitHub Actions currently uses Node.js 14, which will cause the operation to fail.

https://github.com/u3u/prettier-vscode/actions/runs/5269104746/jobs/9526747643

https://github.com/prettier/prettier-vscode/blob/03bbcc6a46e5918cbb662ec833b7bea4d6ee9b85/.github/workflows/main.yml#L13-L20

ntotten:
Looks pretty good, there is one test failing though. Can you take a look?

u3u:
@ntotten Hi, this is a bit strange. All tests passed before you merged the `main` branch, and after I updated it, I ran the tests locally again and they also passed.

**All passed**: https://github.com/prettier/prettier-vscode/actions/runs/5302709062/jobs/9600483439
Partial passed: https://github.com/prettier/prettier-vscode/actions/runs/5304290262/jobs/9601182513

And now there are strange errors in the `main` branch testing, which makes me have to suspect whether it is caused by this submission 🤔 (https://github.com/prettier/prettier-vscode/commit/822ad24288650fdb841c3d9b449d8984118bf3dd)
https://github.com/prettier/prettier-vscode/actions/runs/5304286381/jobs/9600500159

u3u:
@ntotten Hi, I found that this test has been intermittently successful and occasionally failed from the beginning. I don't know what caused this either.

https://github.com/prettier/prettier-vscode/actions/runs/4728089713/jobs/8389302572
https://github.com/prettier/prettier-vscode/actions/runs/4784850492/jobs/8506851252
https://github.com/prettier/prettier-vscode/actions/runs/4784808584/jobs/8506754151
https://github.com/prettier/prettier-vscode/actions/runs/4773168621/jobs/8486085806


image

u3u:
I don't understand why using pnpm 7 would affect this test.
There is no problem testing pnpm 8 on non-Windows systems now.

image

u3u:
https://github.com/u3u/prettier-vscode/actions/runs/5330997873
All tests have passed now. Since Windows cannot use Node.js 18 for testing in CI, only pnpm 7 can be used.

To pass the "it formats by Prettier v3" test when using pnpm 7, `lockfileVersion` of `pnpm-lock.yaml` needs to be restored to `5.4`. I don't know why it is necessary. For future compatibility, I kept the file `.npmrc` under `test-fixtures/plugins-pnpm/`.

sosukesuzuki (approved):

Issue #3017, opened 2023-06-08
Doesn't work with non-node runtimes


cdmistman:
### Summary

In a [`bun`](https://bun.sh) project, I don't have `node` available in the environment (as I'm using nix to manage my developer environment). This extension seems to invoke the `prettier` bin directly, which results in the binary failing to run due to the shebang line `#!/usr/bin/env node`. While `prettier` should be runtime-agnostic (but understandably probably can't fix this), there should be a setting in the extension that allows you to override the runtime used to run `prettier` eg `bun ./node_modules/.bin/prettier`.

### Github Repository to Reproduce Issue

https://github.com/cdmistman/bun-plugin-svelte

### Steps To Reproduce:

1. Initialize a Bun project using `bun init`
2. `bun add prettier -d`
3. Make `.prettierrc` file
4. Make some changes to `index.ts` that violate your `.prettierrc`
5. Run this extension's formatter

### Expected result

The file is formatted

### Actual result

The file is not formatted

### Additional information

_Feel free to attach a screenshot_.

**VS Code Version**:

1.78

**Prettier Extension Version**:

9.13.0

**OS and version**:

MacOS 13.1

## Prettier Log Output

```
["INFO" - 11:47:12 AM] Extension Name: esbenp.prettier-vscode.
["INFO" - 11:47:12 AM] Extension Version: 9.13.0.

```

heartoutx:
I got this error when updating to vscode 1.79

```
["INFO" - 2:00:06 AM] Attempted to determine module path from package.json
["ERROR" - 2:00:06 AM] Failed to load module. If you have prettier or plugins referenced in package.json, ensure you have run `npm install`
["ERROR" - 2:00:06 AM] Cannot read properties of undefined (reading 'uid')
TypeError: Cannot read properties of undefined (reading 'uid')
	at Object.statSync (/home/eric/.vscode-server/bin/b380da4ef1ee00e224a15c1d4d9793e27c2b6302/node_modules/graceful-fs/polyfills.js:313:17)
	at p (/home/eric/.vscode-server/extensions/esbenp.prettier-vscode-9.13.0/dist/extension.js:1:35221)
	at L (/home/eric/.vscode-server/extensions/esbenp.prettier-vscode-9.13.0/dist/extension.js:1:37052)
	at /home/eric/.vscode-server/extensions/esbenp.prettier-vscode-9.13.0/dist/extension.js:1:36677
[...]

Issue #2857, opened 2023-02-02
Run with system node


tjx666:
**Is your feature request related to a problem? Please describe.**

Like ESLint extension provide setting: `"eslint.runtime": "node",`, more details ref: https://github.com/microsoft/vscode-eslint/issues/1233

**Describe the solution you'd like**

Provide setting `prettier.runtime`, and support pass `node` to auto detect system node path.

izaakschroeder:
@tjx666 I have opened #3061 if you want to try testing out that branch. No promises, but it's a start. My first time mucking about in this codebase and VSCode extensions in general… so it would be very useful to have feedback from someone who is actually affected by this problem.

PR #3061, opened 2023-07-03
feat: allow prettier to use custom `node`

izaakschroeder:
Adds a `prettier.runtime` configuration option that, when used, invokes a version of `PrettierWorkerInstance` that uses `fork` instead of worker threads.

Maybe fixes: https://github.com/prettier/prettier-vscode/issues/3017
Fixes: https://github.com/prettier/prettier-vscode/issues/2857

TODO:
- [ ] Run tests – I have a few tests left to fix (not entirely sure why… since the new codepath should only ever be enabled when `prettier.runtime` is set…), but I'm happy for early feedback if this is something the maintainers are willing to consider before moving ahead.

### DIFF
--- a/package.nls.zh-tw.json
+++ b/package.nls.zh-tw.json
@@ -24,6 +24,7 @@
   "ext.config.requireConfig": "排版需要 prettier 組態檔。參閱[可用的組態檔文件](https://prettier.io/docs/en/configuration.html).\n\n> _注意，無標題檔案仍會使用 VS Code 的 prettier 設定進行排版，不受這個設定值影響。_",
   "ext.config.requirePragma": "Prettier 可以限制它自己只對包含特殊註解的檔案進行排版，這個特殊的註解成為 pragma ，位於檔案的最頂部。這對於想要對那些大型、未經過排版的程式碼緩步採納 prettier 非常有幫助。",
   "ext.config.resolveGlobalModules": "這個套件會在區域的模組找不到 prettier 模組時嘗試使用去全域的 npm 或 yarn 模組中尋找。\n> _這個設定會導致效能的負面影響，特別在 Windows 中有掛載網路磁碟機。只有在你必須使用全域模組的情況下再啟用。_",
+  "ext.config.runtime": "TODO TRANSLATE ME",
   "ext.config.withNodeModules": "這個套件會對 node_modules 的檔案進行排版。",
   "ext.config.semi": "是否要在每一列的結尾加上分號。",
   "ext.config.singleQuote": "會使用單引號而非雙引號。",
--- a/src/ChildProcessWorker.ts
+++ b/src/ChildProcessWorker.ts
@@ -2,12 +2,44 @@ import { fileURLToPath, URL } from "url";
 import { ChildProcess, fork, ForkOptions } from "child_process";
 import { EventEmitter } from "events";
 
+const alive = (child: ChildProcess): boolean => {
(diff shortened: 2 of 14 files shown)


src/ChildProcessWorker.ts
```diff
@@ -0,0 +1,105 @@
+import { fileURLToPath, URL } from "url";
+import { ChildProcess, fork, ForkOptions } from "child_process";
+import { EventEmitter } from "events";
+
+export class ChildProcessWorker {
```
izaakschroeder, line 5: This class mimics `Worker` (with `on("message", ...)` and `postMessage(...)`) except it uses `fork` and node's IPC instead.

src/PrettierWorkerInstance.ts
```diff
@@ -34,12 +39,19 @@ export const PrettierWorkerInstance: PrettierInstanceConstructor = class Prettie
     }
   > = new Map();
 
+  private worker;
```
izaakschroeder, line 42: Is there any problem with having a non-global worker instance?

src/worker/prettier-instance-worker.js
```diff
@@ -14,81 +12,84 @@ function requireInstance(modulePath) {
...
+  if (methodName === "getSupportInfo") {
+    result = { languages: result.languages };
+  }
+  return {
+    type: "callMethod",
+    payload: { result, id, isError: false },
+  };
+}
+
+module.exports = (parentPort) => {
```
izaakschroeder, line 26: This function is basically just a wrapped version of the existing message handlers that were already in the class, with a small DRY'd up `normalizeResult` utility to avoid some existing code duplication.

src/PrettierWorkerInstance.ts
```diff
@@ -60,13 +72,16 @@ export const PrettierWorkerInstance: PrettierInstanceConstructor = class Prettie
         }
       }
     });
+    this.worker.on("error", (err) => {
+      context.loggingService.logInfo(`Worker error ${err.message}`, err.stack);
```
izaakschroeder, line 76: Consider: Should the worker be restarted? Seems like a sensible safeguard.

izaakschroeder: Worker is now restarted on crash. Should probably be debouced.

package.nls.zh-tw.json
```diff
@@ -24,6 +24,7 @@
   "ext.config.requireConfig": "排版需要 prettier 組態檔。參閱[可用的組態檔文件](https://prettier.io/docs/en/configuration.html).\n\n> _注意，無標題檔案仍會使用 VS Code 的 prettier 設定進行排版，不受這個設定值影響。_",
   "ext.config.requirePragma": "Prettier 可以限制它自己只對包含特殊註解的檔案進行排版，這個特殊的註解成為 pragma ，位於檔案的最頂部。這對於想要對那些大型、未經過排版的程式碼緩步採納 prettier 非常有幫助。",
   "ext.config.resolveGlobalModules": "這個套件會在區域的模組找不到 prettier 模組時嘗試使用去全域的 npm 或 yarn 模組中尋找。\n> _這個設定會導致效能的負面影響，特別在 Windows 中有掛載網路磁碟機。只有在你必須使用全域模組的情況下再啟用。_",
+  "ext.config.runtime": "TODO TRANSLATE ME",
```
dmaskasky, line 27: 一切都翻译我

Issue #3045, opened 2023-06-28
Regression on package.json finding behavior for globally installed prettier with vscode extension


wxf-contentgroove:
### Summary

The prettier vscode extension seems to have had a behavior regression in version 9.16.0. It is no longer able to load the package.json for a globally installed prettier on macOS

### Steps To Reproduce:

1.  Set prettier path to the globally installed version of prettier, typically: /usr/local/lib/node_modules/prettier on macOS
1.  Try to format a file with the "Format document" command

### Expected result

The file is formatted with prettier

### Actual result

The document is not formatted and the prettier extension throws an error. 

### Additional information

Behavior works as expected on extension version 9.15.0. Additionally, prettier is present in the path and can be run as expected from the command line. 

I have not checked if this behavior regression is also present on Windows or Linux. 

EDIT: as per [this reply](https://github.com/prettier/prettier-vscode/issues/3045#issuecomment-1612982190) to the issue, it appears to also effect windows

Here's the ls of the prettier install location ( /usr/local/lib/node_modules/prettier ) 

```
➜  ~ ls  /usr/local/lib/node_modules/prettier
LICENSE              parser-angular.js    parser-meriyah.js
README.md            parser-babel.js      parser-postcss.js
bin-prettier.js      parser-espree.js     parser-typescript.js
cli.js               parser-flow.js       parser-yaml.js
doc.js               parser-glimmer.js    standalone.js
esm                  parser-graphql.js    third-party.js
index.js             parser-html.js
package.json         parser-markdown.js
```

All the correct files appear to be present

**VS Code Version**:

Version: 1.77.3
Commit: 704ed70d4fd1c6bd6342c436f1ede30d1cff4710
Date: 2023-04-12T09:19:37.325Z (2 mos ago)
Electron: 19.1.11
Chromium: 102.0.5005.196
Node.js: 16.14.2
V8: 10.2.154.26-electron.0
OS: Darwin x64 22.3.0
Sandboxed: No

**Prettier Extension Version**:

v9.16.0

**OS and version**:
macOS Ventura Version 13.2.1 (22D68)

## Prettier Log Output

```
["DEBUG" - 3:22:59 PM] Local prettier module path: '/usr/local/lib/node_modules/prettier'
["INFO" - 3:22:59 PM] Attempted to load Prettier module from /usr/local/lib/node_modules/prettier
[...]

PR #3042, opened 2023-06-23
Fix module resolution with prettierPath setting

vsiao:
Fixes #3041.

#3038 attempts to find a `package.json` starting at a given `modulePath`, but starts one directory further up than it should.

PR #3272, opened 2024-01-24
Fix module resolution with `prettierPath` setting

cesaryuan:
- [x] Run tests

This fix was once proposed in https://github.com/prettier/prettier-vscode/commit/d04b294a15f9dc20e1dc061013b57c7aa8f2764c, but was reverted in https://github.com/prettier/prettier-vscode/commit/f6ff18b500858b913ef43dbad73ecbae929c.

However, this is actually a misunderstanding. Let me analyze the entire process of the matter.
1. #3041 This bug was first raised. The version of Prettier in the issue is 9.16.0
2. #3042 This bug was subsequently fixed, and the fix was first released in version 9.17.0
3. #3045 raised this bug again. The version of Prettier in the issue is also 9.16.0. However, at this time @ntotten [mistakenly thought that this was caused by the change](https://github.com/prettier/prettier-vscode/issues/3045#issuecomment-1614745069) https://github.com/prettier/prettier-vscode/commit/d04b294a15f9dc20e1dc061013b57c7aa8f2764c, but this is obviously impossible, because https://github.com/prettier/prettier-vscode/commit/d04b294a15f9dc20e1dc061013b57c7aa8f2764c did not appear in the release version until 9.17.0. (I don’t mean to blame ntotten)
4. At the same time, we can see in #3045 and #3041 that after the issue was closed, [most people reported that the bug had not been fixed](https://github.com/prettier/prettier-vscode/issues/3045#issuecomment-1614994507), which further illustrates what was reverted in https://github.com/prettier/prettier-vscode/commit/f6ff18b500858b913ef43dbad73ecbae929c3b27 is wrong.

Therefore, I am re-fixing this long-standing bug today. 🥰

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 ## [Unreleased]
 
-Fixed the issue where VSCode was misrecognizing the path in output panel due to added quotes.
+- Fixed the issue where VSCode was misrecognizing the path in output panel due to added quotes.
+- Fix module resolution with `prettierPath` setting
 
 ## [10.1.0]
 
--- a/src/ModuleResolver.ts
+++ b/src/ModuleResolver.ts
@@ -114,7 +114,7 @@ export class ModuleResolver implements ModuleResolverInterface {
           return pkgFilePath;
         }
       },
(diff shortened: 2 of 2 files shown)


ntotten:
Awesome! Thanks for digging into this. 🙏

I will try to merge and ship this soon.

PR #3346, opened 2024-03-21
Updated to prettier 3

ntotten:
- [ ] Run tests


src/BrowserModuleResolver.SupprotInfo.ts
Stanzilla, line 1: typo in file name

astorije:
Hi there! I'm very interested to see this happen (and happy to help if I can), would you mind explaining what is blocking this?

Lycolia:
I think it's probably because the test is failing and there's a typo in the code. The typo can be fixed quickly, but fixing this test seems difficult, as it seems to be caused by changes in the build ecosystem.

Specifically, it seems that the plugin is no longer loaded correctly when the test is run, and the format is executed without the plugin, which is why the test is failing.

Issue #3424, opened 2024-06-06
Wrong .prettierignore file used when using multiple sub folders


Kamahl19:
### Summary

I have a monorepo project with multiple packages. Each of the them contains a .prettierignore and .prettierrc. Plus one at the root of the project for files located in the root. When I'm saving a file, a .prettierrc config file which is located closest to the saved file is being used. This is correct behaviour. However this does not work correctly for the .prettierignore file. Only the .prettierirnogr in project root is always used, no matter which file is being saved. Instead the .prettierignore closest to the saved file should be used.

This is basically a follow-up of the issue (https://github.com/prettier/prettier-vscode/issues/3003) already closed by the bot.

### Github Repository to Reproduce Issue

I am using this repo https://github.com/KirianCaumes/Prettier-VsCode-Bug which was already used in the duplicate issue https://github.com/prettier/prettier-vscode/issues/3003 which was closed by the bot.

### Steps To Reproduce:

1.  Open the ./should-be-ignored.yml file, and save it. It should be skipped because of the ./.prettierignore.
1.  Open the ./front/should-not-be-ignored.yml file, and save it. It is wrongly skipped because the ./.prettierignore is being used. Instead, the ./front/.prettierignore should be used the file should not be skipped. 

### Expected result

When saving a file in a subfolder with it's own .prettierignore, this one should be used, not the one at the root of the project. The behaviour should be the same as for the .prettierrc config.

### Actual result

The root .prettierignore is used instead of the closest one.

### Additional information

**VS Code Version**:

Version: 1.89.1
Commit: dc96b837cf6bb4af9cd736aa3af08cf8279f7685
Date: 2024-05-07T05:14:32.757Z
Electron: 28.2.8
ElectronBuildId: 27744544
Chromium: 120.0.6099.291
Node.js: 18.18.2
V8: 12.0.267.19-electron.0
OS: Darwin arm64 23.5.0

**Prettier Extension Version**:

v10.4.0

**OS and version**:

MacOS 14.5

StanlieK:
I have the same issue

Kamahl19:
@ntotten I am sorry to bother you but this issue has been raised twice (https://github.com/prettier/prettier-vscode/issues/3003) and always closed by bot. Would you please comment whether this is something you plan to fix or not at all? Monorepos are suffering by this. Thanks a lot!

Simon7520:
I have the same problem

Issue #3003, opened 2023-05-11
Wrong `.prettierignore` file used when using multiple sub folders


KirianCaumes:
### Summary

I have a project structure with two folders : `front` and `back`.
Each of the them contains a `.prettierignore` and `.prettierrc.json`. Plus one at the root of the project.
When I'm saving a file (via `"editor.formatOnSave": true`), depending if the file is in the root, `front` or `back` folder, the correct `.prettierrc.json` is used, but not the correct `.prettierignore`.

### Github Repository to Reproduce Issue

Here is a repository to reproduce the issue: https://github.com/KirianCaumes/Prettier-VsCode-Bug 

### Steps To Reproduce:

1.  Open the `./should-be-ignored.yml` file, and save it. It should be skiped thanks to `./.prettierignore`.
1.  Open the `./front/should-not-be-ignored.ymll` file, and save it. It is skiped thanks to `./.prettierignore`, but it should not because of the `./front/.prettierignore`. 

### Expected result

When saving a file in a subfolder with it's own `.prettierignore`, this one should be used, not the one at the root of the project.
The behavior I want, is exactly how it's working now with `.prettierrc.json`.

### Actual result

The root `.prettierignore` is used instead of the closest one.

### Additional information

**VS Code Version**:

Version: 1.78.1 (user setup)
Commit: 6a995c4f4cc2ced6e3237749973982e751cb0bf9
Date: 2023-05-04T09:48:08.683Z
Electron: 22.5.1
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.22000
Sandboxed: No

**Prettier Extension Version**: 

v0.50.0

**OS and version**:

Windows 11

## Prettier Log Output

When saving `should-be-ignored.yml`:

```
["INFO" - 4:06:57 PM] Formatting file:///home/project/should-be-ignored.yml
["INFO" - 4:06:57 PM] Using config file at '/home/project/.prettierrc.json'
["INFO" - 4:06:57 PM] Using ignore file (if present) at /home/project/.prettierignore
["INFO" - 4:06:57 PM] File Info:
{
  "ignored": true,
  "inferredParser": null
}
["INFO" - 4:06:57 PM] File is ignored, skipping.
```

When saving `front/should-not-be-ignored.yml`, we clearly see that the `.prettierrc.json` is the one from the `front` folder, but not the `.prettierignore`:

```
[...]

PR #3622, opened 2025-02-10
Fix issue that .prettierignore in monorepo sub folders is not read. (#3003, #3424)

AdoKevin:
- [x] Run tests

This issue has been metion in thread  #3003 and #3424. Currently, Prettier only reads .pretterignore file from directories included by *.code-workspace* or root project directory. However, in monorepo cases, we prefer no *.code-workspace* is created, and packages are organized by subdictories.

My main change is to prioritize reading the .prettierignore file alongside the .prettierrc file (I believe is the most common usage). If it does not exist, the current logic will still be attempted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](https://keepachangelog.com/) for recommendations on how to structure this file. -->
 
+## [Unreleased]
+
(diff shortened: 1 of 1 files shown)


JackRobards:
When you have some time @ntotten, do you think it would be possible to consider this PR, or did you have any concerns with this change?

Kamahl19:
@ntotten this PR makes the extension so much usable in monorepos

Kamahl19:
@ntotten Hey, I believe this PR was overlooked and therefor auto-closed but it brings huge value to monorepos. Without this we need to duplicate all ignores both in the root .prettierignore and in nested .prettierignore files.

Issue #2828, opened 2022-12-29
Feature Request: settings.json entry for plugins


ChadBailey:
**Is your feature request related to a problem? Please describe.**
When using prettier with non-javascript languages, it requires the setup of the plugin within the project's `package.json` as the docs indicate [is required for plugin support](https://github.com/prettier/prettier-vscode#plugins). This makes sense when using prettier as a node module executed from the command line, but that doesn't make a lot of sense when using Prettier from the context of a visual studio extension. This problem is exasperated by the fact if you are intending to use a plugin, you likely aren't using JavaScript for that project.

**Describe the solution you'd like**
I propose adding the following settings.json entry: `"prettier.plugins":["@prettier/plugin-ruby","prettier-plugin-toml"]` such that the array values are strings that match the string format expected by `npm install` - see this [plugin list](https://prettier.io/docs/en/plugins.html#official-plugins) for the plugins I recommend being officially supported (but it could be designed in a generic way)

**Describe alternatives you've considered**

- Installing plugins via `npm install -g <name>` then have prettier auto-discover and use them. Rejected because user may not have need or want node/npm installed globally just to autoformat the language of their choosing.

**Additional context**

In an ideal world, this would be accompanied with logic to identify when the user is attempting to autoformat a document in a language supported by a plugin. If they didn't have the plugin the user would be prompted to download and add the plugin to their config rather than receiving this error

![image](https://user-images.githubusercontent.com/15102816/210007378-eccbad15-e9e5-43d1-804a-090e29dc8d6c.png)

ChadBailey:
> Effectively something like recommended plugins but for modules.

I'm not entirely sure you're fully understanding what I'm asking based on this (and by all means I could be the one confused here). I realize I didn't do my best job at making my request clear, so I'm going to rewrite the problem statement, here goes...

## Problem Statement

When using prettier visual studio code extension for formatting any language supported through a 3rd party plugin, I must do that by installing node, configuring npm, and putting a `package.json` file in my project root. (Unless I'm misunderstanding the docs).

This is an issue for the following reasons:
- Project directory gets polluted with extension-specific files
- The files are not a dependency of the project, but of the extension
- Other developers on the same project may not want to use prettier, and may require a diverging `package.json` and `package-lock.json`
- Requires what might be an otherwise totally unnecessary installation of node and configuration of npm.
- Will need to be repeated for every single project, can't be globally configured

[...]

PR #3782, opened 2025-09-23
Add support for configuring global Prettier plugins via prettier.plugins

ayloncarrijo:
This PR introduces a new `prettier.plugins` setting in `settings.json` that allows for the automatic download and use of global Prettier plugins. This makes it easier to use Prettier with languages it doesn't support out of the box, without needing a `package.json` or local npm setup. It also simplifies quick one-off edits, allowing you to open a file, format it, and move on without additional configuration.

See #2828

- [x] Run tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to the "prettier-vscode" extension will be documented in thi
 
 <!-- Check [Keep a Changelog](https://keepachangelog.com/) for recommendations on how to structure this file. -->
 
+## [Unreleased]
+
+- Add support for global Prettier plugins via `prettier.plugins` setting
+
 ## [11.0.0]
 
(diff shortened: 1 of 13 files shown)


roele:
I tried installing a build of this PR but could not getting it to work properly. Plugins did not resolve as defined in the `.prettierrc` file for some reason.

What i am looking for is a way to use globally installed plugins. For example for non JS projects where we don't want a `node_modules` folder or `package.json`.

Ideally we could install prettier and plugins globally and then simply use them in the `.prettierrc` file. So far i could not get this to work properly. I can define the `prettier.prettierPath` to point to my globally installed `prettier` but the plugins are not resolved. Event with `prettier.resolveGlobalModules` set it does not load the plugins. 

One workaround i found is to define the absolute paths to the plugins but that is ugly and does not work well across different OSes.
```yaml
plugins:
  - "/Users/foo/.local/share/mise/installs/node/latest/lib/node_modules/prettier-plugin-java/dist/index.js"
```

